PR #93, opened 2020-04-02
 support multiple public keys 

penpyt:
# Description

https://scalar-labs.atlassian.net/browse/DLT-5594

## Tasks:
- [x] copy the public key inside authorized key file on every server
- [x] support azure
- [x] support aws
- [ ] update docs


feeblefakie:
@penpyt CI fails. Please check !

modules/universal/bastion/main.tf
```diff
@@ -15,6 +15,10 @@ resource "null_resource" "ansible_playbooks_copy" {
     private_key = file(var.private_key_path)
   }
 
+  provisioner "local-exec" {
+    command = "for i in ${var.multiple_public_key_folder_path}/*.pub; do if [[ -e $i ]]; then mkdir -p ../../../provision/ansible/playbooks/files/ssh_public; cp $i ../../../provision/ansible/playbooks/files/ssh_public/; fi; done"
```
feeblefakie, line 19: I think we can make it multiple lines for ease of read ?

provision/ansible/playbooks/bastion-server.yml
```diff
@@ -10,3 +10,21 @@
...
+  tasks:
+    - name: test if ssh_public folder exist
+      local_action: stat path="files/ssh_public"
+      register: folder_details
+
+    - name: add key
+      authorized_key:
+        user: centos
+        state: present
+        key: "{% for key in lookup('fileglob', 'files/ssh_public/*.pub').split(',') %}{{ lookup('file', key) ~ '\n'}}{% endfor %}"
```
feeblefakie, line 26: Is it OK to glob with `.pub` suffix ?
I think public keys may not have `.pub` suffix.

examples/aws/network/example.tfvars
```diff
@@ -10,12 +10,14 @@ private_key_path = "./example_key"
 
 internal_domain = "internal.scalar-labs.com"
 
+multiple_public_key_folder_path = "./public_key"
```
feeblefakie, line 13:
```suggestion
public_key_folder_path = "./public_key"
```
since it does't have to have multiple keys.
Also, we should place it close to `public_key_path` since it is a similar setting.

feeblefakie, line 13: What will happen if `public_key_path` and `public_key_folder_path` are both specified ?
Both keys are added ?

penpyt: Yes agree, will refactor

penpyt: if the keys in public_key_path and  public_key_folder_path  are different, both will be added.
if not Ansible will only check the key on the remote server.

penpyt: changed in https://github.com/scalar-labs/scalar-terraform/pull/93/commits/d0b2c6244ef2b6fa1caa9319f404f22e5b4f97f1

modules/universal/bastion/main.tf
penpyt: updated in https://github.com/scalar-labs/scalar-terraform/pull/93/commits/5d0766362b04c2fba4d7831c310ccd5055e60551

provision/ansible/playbooks/bastion-server.yml
penpyt: agree, fixed in https://github.com/scalar-labs/scalar-terraform/pull/93/commits/a1e4442f1499104f64816a3a0d7842524fed4b9a

PR #97, opened 2020-04-06
Add C* username and password to scalardl and reaper modules

tei-k:
# Description

https://scalar-labs.atlassian.net/browse/DLT-5701


modules/universal/scalardl/provision/Dockerfile
```diff
@@ -16,13 +16,6 @@ COPY ./create_schema.sh ./
 RUN chmod 755 ./startup.sh
 RUN chmod 755 ./create_schema.sh
 
-ENV SCALAR_CASSANDRA_HOST ${SCALAR_CASSANDRA_HOST}
-ENV SCALAR_CASSANDRA_PORT 9042
-ENV SCALAR_CASSANDRA_USERNAME ''
-ENV SCALAR_CASSANDRA_PASSWORD ''
-ENV CASSANDRA_REPLICATION_FACTOR ${CASSANDRA_REPLICATION_FACTOR}
-
-
```
ymorimo, line 25: I removed these ENV instructions because they don’t affect the resulting image:

```console
$ docker inspect scalar-ledger-provision:2.0.3
...
            "Env": [
                "SCALAR_CASSANDRA_HOST=cassandra",
                "SCALAR_CASSANDRA_PORT=9042",
                "SCALAR_CASSANDRA_USERNAME=",
                "SCALAR_CASSANDRA_PASSWORD=",
...
```
The variables come from the original `scalarlabs/scalar-ledger` image.

To set build-time env variables in Dockerfile, it was needed to pass them from `build` > `args` in docker-compose.yml, but since we don't have them, `${}`s in this Dockerfile have not really been substituted. Moreover it is not a good idea to store the concrete credentials in the image itself.
The configuration variables are passed at runtime from docker-compose.yml, so there's no problem.

Here is the current docker-compose.yml:
```yml
services:
  scalar:
    build:
      context: ./
      args:
        - FROM_SCALAR_IMAGE=${FROM_SCALAR_IMAGE}
    image: ${SCALAR_IMAGE}
    environment:
      - SCALAR_CASSANDRA_HOST=${SCALAR_CASSANDRA_HOST}
      - SCALAR_CASSANDRA_USERNAME=${SCALAR_CASSANDRA_USERNAME}
      - SCALAR_CASSANDRA_PASSWORD=${SCALAR_CASSANDRA_PASSWORD}
      ...
```

ymorimo:
@tei-k It seems that I can't add you to the reviewers because you're the creator of this PR. Please review!

modules/universal/reaper/provision/docker-compose.yml
```diff
@@ -0,0 +1,31 @@
...
+    image: scalar-cassandra-reaper:latest
+    build:
+      context: ./
+    command:
+     - "cqlsh"
+     - "--cqlversion=3.4.4"
+     - "--username=${REAPER_CASS_AUTH_USERNAME}"
+     - "--password=${REAPER_CASS_AUTH_PASSWORD}"
+     - "--execute"
+     - "CREATE KEYSPACE reaper_db WITH replication = {'class': 'SimpleStrategy', 'replication_factor': ${CASSANDRA_REPLICATION_FACTOR} };"
```
feeblefakie, line 13:
```suggestion
     - "CREATE KEYSPACE IF NOT EXISTS reaper_db WITH replication = {'class': 'SimpleStrategy', 'replication_factor': ${CASSANDRA_REPLICATION_FACTOR} };"
```

I did a quick check and `CREATE KEYSPACE` returns 2 if there is the keyspace already. When specified with `IF NOT EXISTS`, it returns 0 even if there is the keyspace.

I am wondering if it will keep restarting the container without `IF NOT EXISTS` when C* password is changed.

modules/universal/scalardl/provision/Dockerfile
feeblefakie: Should I update [this](https://github.com/scalar-labs/scalar/blob/master/ledger/Dockerfile) as well ?

modules/universal/reaper/provision/docker-compose.yml
```diff
@@ -0,0 +1,31 @@
+version: '3.7'
```
feeblefakie, line 1: Just a question.
So, when we update C* password after C* is created, we need to restart reaper container manually right ?

tei-k:
LGTM! 👍

modules/universal/scalardl/provision/Dockerfile
ymorimo: I will consider we should update it as well.

modules/universal/reaper/provision/docker-compose.yml
ymorimo: @feeblefakie Right. I noticed that the cqlsh container that runs `CREATE KEYSPACE` keeps restarting when it is run second time, so I was going to report a separate issue. Simply adding `IF NOT EXISTS` should solve it, should I apply the fix here?

feeblefakie: It can be separated but I think it's fine to apply this here.

modules/universal/reaper/provision/docker-compose.yml
ymorimo: Tainting the `null_resource` updates the username and the password in `env` file but the container should not be restarted automatically as long as the image is the same. I'll check on that quickly.

ymorimo: It seems that changing environment variables lets the `docker-compose up` command recreate the container. So when you taint `null_resource.reaper_waitfor` and apply, you don't need to restart the container manually.

```console
$ docker-compose up -d
provision_reaper_1 is up-to-date
# edit env
$ source ./env
$ docker-compose up -d
Recreating provision_reaper_1 ... done
$ docker-compose up -d
provision_reaper_1 is up-to-date
```

modules/universal/scalardl/provision/Dockerfile
ymorimo: @feeblefakie In scalar repository we should not remove `ENV`s from Dockerfile since they are default values.

PR #98, opened 2020-04-06
Multi az for aws

tei-k:
# Description

https://scalar-labs.atlassian.net/browse/DLT-5564

# Done
- Support for Multi-AZ.
- Create nat gateway each az.
- Add `Network` tag for vpc.
- Some refactoring.
- Overwrite `create_schema.cql.tmpl` for test (need revert)


modules/aws/network/README.md
```diff
@@ -24,7 +24,7 @@ The Network module creates a virtual network with subnets.
 | image_id | The image id to initiate. |
 | internal_domain | The internal domain for setting srv record |
 | key_name | The key-name of the AWS managed ssh key_pair. |
-| location | The AWS availability zone to deploy environment. |
+| locations | The AWS availability zone to deploy environment. |
```
feeblefakie, line 27:
```suggestion
| locations | The AWS availability zones to deploy environment. |
```

feeblefakie:
@yito88 Please review especially for C* part .

examples/aws/network/example.tfvars
```diff
@@ -14,7 +18,8 @@ internal_domain = "internal.scalar-labs.com"
 
 network = {
   # bastion_resource_type     = "t3.micro"
-  # bastion_resource_count    = "1"
+  # bastion_resource_count    = "2"
+  # bastion_dns_index         = "1"
```
feeblefakie, line 22: I feel we don't need this variable and making the system always prefer smaller index is fine.

modules/aws/network/output.tf
```diff
@@ -87,7 +87,7 @@ UserKnownHostsFile /dev/null
 StrictHostKeyChecking no
 
 Host bastion
-HostName ${module.bastion.bastion_host_ips[0]}
+HostName ${module.bastion.bastion_host_ips[local.network.bastion_dns_index - 1]}
```
feeblefakie, line 90: I feel hard coding `0` is fine. In case of failure, we can manually change the ssh.cfg to specify another bastion host because updating terraform file and doing `apply` for getting new `ssh.cfg` might be too much. 
Thought ?

modules/aws/cassandra/cassandra.tf
```diff
@@ -42,7 +42,7 @@ module "cassandra_cluster" {
 resource "aws_ebs_volume" "cassandra_data_volume" {
   count = local.cassandra.enable_data_volume && ! local.cassandra.data_use_local_volume ? local.cassandra.resource_count : 0
 
-  availability_zone = local.location
+  availability_zone = local.locations[count.index]
```
yito88, line 45: `count.index` can be out of index?

modules/aws/cassandra/cassandra.tf
```diff
@@ -113,7 +113,7 @@ resource "null_resource" "volume_data_local" {
 resource "aws_ebs_volume" "cassandra_commitlog_volume" {
   count = local.cassandra.enable_commitlog_volume && ! local.cassandra.commitlog_use_local_volume ? local.cassandra.resource_count : 0
 
-  availability_zone = local.location
+  availability_zone = local.locations[count.index]
```
yito88, line 116: `count.index`can be out of index?

modules/aws/cassandra/cassandra.tf
tei-k: 👍 😱
Fixed in 5b96e09.

modules/aws/cassandra/cassandra.tf
tei-k: ditto. (5b96e09)

modules/aws/cassandra/cassandra.tf
feeblefakie: 😱 
Good catch ! > @yito88 

@tei-k please check with more than 3 nodes. (say 6 nodes)

modules/aws/network/README.md
tei-k: Fixed in e2df875

examples/aws/network/example.tfvars
tei-k: 👌 Fixed in e2df875.

modules/aws/network/output.tf
tei-k: 👍 Fixed in e2df875.

Maybe better to output all the IPs of bastion hosts. 
But we can get it from web console, so it's not required. 😅

modules/aws/cassandra/cassandra.tf
tei-k: I checked it with `enable_data_volume=true` and `enable_commitlog_volume=true`.
It looks like it's working fine.

examples/aws/monitor/example.tfvars
```diff
@@ -6,6 +6,7 @@ monitor = {
   # resource_type             = "t3.small"
   # resource_root_volume_size = "64"
   # resource_count            = "1"
+  # dns_index                 = "1"
```
feeblefakie, line 9: I think `dns_index` is too internal implementation specific (dns_index is used for the index of DNS entries) so it's a bit hard for users to assume what it is.
Maybe making it `active_offset` and 0-based is better ?

tei-k: >Maybe making it active_offset and 0-based is better ?

👍 Fixed in d1a01b7.

feeblefakie:
@penpyt Please take a look. It should be checked with as many as eyes possible.

test/modules/awsdeploy/network/main.tf
```diff
@@ -3,7 +3,7 @@ module "network" {
 
   # Required Variables
   name                        = var.name
-  location                    = var.location
+  locations                   = var.locations
```
feeblefakie, line 6: Is it possible to throw error if the number of locations is more than 3 ?

modules/universal/cassandra/main.tf
```diff
@@ -42,7 +42,7 @@ resource "null_resource" "cassandra" {
     inline = [
       "cd ${module.ansible.remote_playbook_path}/playbooks",
       "echo '${var.cassy_public_key}' > cassandra.pub",
-      "ansible-playbook -u ${var.user_name} -i ${var.host_list[count.index]}, cassandra-server.yml -e CASSANDRA_MEMTABLE_THRESHOLD=${var.memtable_threshold} -e CASSANDRA_SEEDS=${join(",", var.host_seed_list)} -e enable_tdagent=${var.enable_tdagent ? 1 : 0} -e CASSANDRA_STATE=${var.start_on_initial_boot ? "started" : "stopped"} -e monitor_host=monitor.${var.internal_domain}",
+      "ansible-playbook -u ${var.user_name} -i ${var.host_list[count.index]}, cassandra-server.yml -e CASSANDRA_MEMTABLE_THRESHOLD=${var.memtable_threshold} -e CASSANDRA_SEEDS=${join(",", var.host_seed_list)} -e enable_tdagent=${var.enable_tdagent ? 1 : 0} -e CASSANDRA_STATE=${var.start_on_initial_boot ? "started" : "stopped"} -e monitor_host=monitor.${var.internal_domain} -e cassandra_rack_index=${count.index % length(var.locations)}",
```
feeblefakie, line 45: Just a question. If cassandra offset 1 is tainted for some reason such as failure and re-applied, newly created cassandra node uses the offset 1, right ?

ymorimo:
It seems that monitor hosts can't be monitored each other with cadvisor when two monitor hosts are deployed.

Screen Shot 2020-04-21 at 14 40 21
Screen Shot 2020-04-21 at 14 40 28

On monitor-2 host:
```console
$ curl http://monitor-1.internal.scalar-labs.com:18080/metrics
curl: (7) Failed connect to monitor-1.internal.scalar-labs.com:18080; Connection timed out
```

test/modules/awsdeploy/network/main.tf
tei-k: Yes, Because always create 3 subnets for public subnet.

https://github.com/terraform-aws-modules/terraform-aws-vpc/blob/23b0a02d2a3225a05ddf5f31d9a32728bf5505c3/main.tf#L888

modules/universal/cassandra/main.tf
tei-k: Right!

provision/ansible/playbooks/roles/cassandra/templates/cassandra-rackdc.properties.j2
```diff
@@ -0,0 +1,3 @@
+dc=dc1
+rack=rack{{ cassandra_rack_index }}
```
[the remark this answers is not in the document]
tei-k: Good point!

Ease of Coding was a priority.
Maybe we don't have `dc2` and `dc3`, so I think it's OK to start from `0`.

tei-k:
>It seems that monitor hosts can't be monitored each other with cadvisor when two monitor hosts are deployed.

Good point!

I suppose it all depends on how well they keep tabs on each other.
We can open a port(18080) and have it monitored, but currently we may get two slack notifications when we create two instances.

ymorimo:
@tei-k I just found that this issue is not related to Multi-AZ, so I don't think it needs to be fixed it in this PR.

tei-k:
I've made a few additional corrections.

- Fix default tag to 2.0.4 for scalardl.
- Fix monitor dns name.
- Delete prometheus dns.
- Delete template file for test.

---

There is a problem with the image size being too large.
```
[centos@scalardl-blue-1 ~]$ docker images
REPOSITORY                TAG                 IMAGE ID            CREATED             SIZE
scalar-ledger-provision   2.0.4               098ab1b4faef        4 minutes ago       846MB
scalar-ledger-provision   2.0.3               835c61ce9b0c        2 days ago          481MB
```

modules/universal/scalardl/provision/Dockerfile
```diff
@@ -1,7 +1,8 @@
-ARG FROM_SCALAR_IMAGE=scalarlabs/scalar-ledger:2.0.3
+ARG FROM_SCALAR_IMAGE=scalarlabs/scalar-ledger:2.0.4
 FROM ${FROM_SCALAR_IMAGE}
 
 #Install cqlsh tool
+RUN mkdir -p /usr/share/man/man1
```
tei-k, line 5: 📝 Beacuse failed In new image. (ref: https://github.com/puckel/docker-airflow/issues/182)

```
update-alternatives: using /usr/lib/jvm/java-8-openjdk-amd64/jre/bin/rmid to provide /usr/bin/rmid (rmid) in auto mode
update-alternatives: error: error creating symbolic link '/usr/share/man/man1/rmid.1.gz.dpkg-tmp': No such file or directory
```

tei-k:
When use `openjdk:8u212-jre-slim-stretch`
```
[centos@scalardl-blue-1 ~]$ docker images
REPOSITORY                TAG                 IMAGE ID            CREATED             SIZE
scalar-ledger-provision   2.0.4               72027c4645b4        9 minutes ago       548MB
scalar-ledger-provision   2.0.3               835c61ce9b0c        2 days ago          481MB
```

feeblefakie:
@tei-k Thanks ! Let's adjust what openjdk to use later.

PR #101, opened 2020-04-14
Fix dns for azure module

tei-k:
# Description
https://scalar-labs.atlassian.net/browse/DLT-5652

# Done
- Add suffix `count.index` to dns.
- Fix to can be create multi bastion.


modules/azure/monitor/main.tf
```diff
@@ -90,26 +90,48 @@ module "monitor_provision" {
   internal_domain               = local.internal_domain
 }
 
-resource "azurerm_private_dns_a_record" "monitor-dns" {
+resource "azurerm_private_dns_a_record" "monitor" {
```
feeblefakie, line 93: How about `monitor-cluster-dns` for monitor.xyz.com and `monitor-host-dns` for monitor-1.xyz.com ?

BTW, all the dns variables are not very consistent. Some uses `-` and others use `_` instead, also some has `a` but the others don't.

feeblefakie: @tei-k Please take a look at the comment !

tei-k: Thanks comment! Changed below.

- Fix resource name to  `monitor_cluster_dns` and `monitor_host_dns`.
- Fix `-` to `_` in resoruce_name.
- Delete prometheus dns name. (doesn't seem necessary)

ymorimo: > Delete prometheus dns name. (doesn't seem necessary)

There is a DNS record resource for prometheus on AWS side too. Let's remove it later.

PR #102, opened 2020-04-14
Add dev for aws

tei-k:
# Description
https://scalar-labs.atlassian.net/browse/DLT-4907


examples/aws/cassandra/example.tfvars
```diff
@@ -1,6 +1,6 @@
 region = "ap-northeast-1"
 
-base = "default" # bai, chiku, sho
+base = "default" # dev, bai, chiku, sho
```
feeblefakie, line 3: The default is bai ?
then 
```suggestion
base = "default" # dev, bai (default), chiku, sho
```
is more user friendly.

modules/aws/cassandra/locals.tf
```diff
@@ -40,6 +40,12 @@ locals {
   cassandra_base = {
     default = local.cassandra_default
 
+    dev = merge(local.cassandra_default,
+      {
+        resource_type = "r5d.large"
```
feeblefakie, line 45: Maybe `t3a.large` is good for dev.

BTW, `t3a` instance type in addition to `t3`. `t3a` is a bit cheaper with the same spec. (newer ?)
scalardl is usign t3.medium, so we might need to reconsider the types later.

modules/aws/scalardl/locals.tf
```diff
@@ -48,6 +48,13 @@ locals {
   scalardl_base = {
     default = local.scalardl_default
 
+    dev = merge(local.scalardl_default,
+      {
+        blue_resource_count  = 2
+        green_resource_count = 2
```
ymorimo, line 54: Do we need two scalardl nodes for both blue and green?

examples/aws/cassandra/example.tfvars
tei-k: `default` is not equal `bai` currently.

Do you want the following?
```
dev, bai (default), chiku, sho
```

modules/aws/cassandra/locals.tf
tei-k: I'm not changed machine spec. 

From requirements.
```
machine spec should be the same as bai
```

examples/aws/cassandra/example.tfvars
feeblefakie: default is chiku ? then `chiku (default)`
 I'm just saying it should be more user friendly since it doesn't say what is the default.

modules/aws/scalardl/locals.tf
tei-k: 🤯 Reverted (8aeaf8a)

examples/aws/cassandra/example.tfvars
tei-k: I think the default is for intergration testing. (from old tools).

It would be better to clarify the definition of default and dev

modules/aws/cassandra/locals.tf
feeblefakie: Sorry. 🙇  I realized that bai is still kind of large for dev.

PR #103, opened 2020-04-14
Make it able to pass S3 bucket name for cassy module

ymorimo:
https://scalar-labs.atlassian.net/browse/DLT-5554

If the S3 bucket name is passed to the cassy arguments for AWS,

* An IAM role for EC2 instances to interact with the S3 bucket is created and attached.
* The S3 bucket name is added to cassy.properties file.


ymorimo:
@tei-k @feeblefakie This is still a draft PR because I'm going to update CassySetup.md later. It would be nice if you could check the code in advance.

examples/aws/cassandra/example.tfvars
```diff
@@ -24,6 +24,7 @@ cassy = {
   #resource_count            = "1"
   #resource_root_volume_size = "64"
   #enable_tdagent            = "true"
+  #s3_bucket_name            = ""
```
feeblefakie, line 27: I think we can use Cassy naming for this since it's a cassy configuration and the name is abstracted.
```suggestion
  #storage_base_uri            = ""
```
And users would specify `s3://bucket-name` for AWS.

modules/aws/cassandra/cassandra.tf
```diff
@@ -12,6 +12,7 @@ module "cassandra_cluster" {
   subnet_id                   = local.subnet_id
   associate_public_ip_address = false
   hostname_prefix             = "cassandra"
+  iam_instance_profile        = "%{if length(aws_iam_instance_profile.cassandra_profile) > 0}${aws_iam_instance_profile.cassandra_profile[0].name}%{endif}"
```
tei-k, line 15: [Q] Wouldn't that be an error when bucket_name is blank?

It could be written as follows.
```
iam_instance_profile = local.cassy.s3_bucket_name != "" ? aws_iam_instance_profile.cassandra_profile[0].name : ""
```

modules/aws/cassandra/cassandra.tf
```diff
@@ -181,6 +182,70 @@ resource "null_resource" "volume_commitlog_local" {
   }
 }
 
+resource "aws_iam_instance_profile" "cassandra_profile" {
```
tei-k, line 185: (IMO) I think don't need `_profile` suffix.

modules/aws/cassandra/cassandra.tf
```diff
@@ -181,6 +182,70 @@ resource "null_resource" "volume_commitlog_local" {
...
+  assume_role_policy = <<EOF
+{
+  "Version": "2012-10-17",
+  "Statement": {
+    "Effect": "Allow",
+    "Principal": { "Service": "ec2.amazonaws.com" },
+    "Action": "sts:AssumeRole"
+  }
+}
+EOF
```
tei-k, line 207: (IMO) Can use `data` resource to define assume role.

e.g.
```
data "aws_iam_policy_document" "assume" {
  statement {
    actions = ["sts:AssumeRole"]

    principals {
      type        = "Service"
      identifiers = ["ec2.amazonaws.com"]
    }
  }
}

resource "aws_iam_role" "cassandra" {
  ・・・
  assume_role_policy = data.aws_iam_policy_document.assume.json
}
```

examples/aws/cassandra/example.tfvars
ymorimo: Good point. But we need an S3 bucket name to create a policy for EC2 instances to interact with S3. This is very AWS specific. Even if we abstract the variable name at this stage, we still need to assume that the value is an `s3://` URI since the bucket name must be extracted from it.
On the other hand [the `universal` module receives an abstracted URI](https://github.com/scalar-labs/scalar-terraform/pull/103/files#diff-84f2d0aebbe5577ab4dc44a45d3a5318R52) like `s3://bucket-name` from this `aws` module.

modules/aws/cassandra/cassandra.tf
ymorimo: 👍 Updated in 43025e4.

modules/aws/cassandra/cassandra.tf
ymorimo: @tei-k What are the advantages of converting the JSON to a data resource?

modules/aws/cassandra/cassandra.tf
ymorimo: It won't be an error because the `if` statement is just a string interpolation so it results in a blank string if the condition is false.

modules/aws/cassandra/cassandra.tf
tei-k: FYI
https://learn.hashicorp.com/terraform/aws/iam-policy#choosing-a-configuration-method
The recommended approach to building AWS IAM policy documents within Terraform is the highly customizable aws_iam_policy_document data source. A short list of benefits over other methods include:
```
tive Terraform configuration - no need to worry about JSON formatting or syntax
Policy layering - create policy documents that combine and/or overwrite other policy documents
Built-in policy error checking
```

modules/aws/cassandra/cassandra.tf
```diff
@@ -181,6 +182,70 @@ resource "null_resource" "volume_commitlog_local" {
...
+        "s3:PutObject",
+        "s3:GetObject",
+        "s3:DeleteObject",
+        "s3:PutObjectAcl"
+      ],
+      "Resource": ["arn:aws:s3:::${local.cassy.s3_bucket_name}/*"]
+    }
+  ]
+}
+EOF
```
tei-k, line 246: Can write the same in data resource.

e.g. (just idea 😇)
```
data "aws_iam_policy_document" "cassandra" {
  version = "2012-10-17"

  statement {
    effect = "Allow"

    actions = [
      "s3:PutObject",
      "s3:GetObject",
      "s3:ListBucket",
      "s3:DeleteObject",
      "s3:PutObjectAcl",
    ]

    resources = [
      "arn:aws:s3:::${local.cassy.s3_bucket_name}",
      "arn:aws:s3:::${local.cassy.s3_bucket_name}/*",
    ]
  }
}
```

examples/aws/cassandra/example.tfvars
tei-k: Considering similar interfaces in AWS and Azure,  `storage_base_uri` might be good.
Maybe we can use `replace` function to get bucket name. (e.g. replace(storage_base_uri, "s3://", ""))

ymorimo:
@tei-k @feeblefakie I have just made the `storage_base_uri` required in a04e693 (Will fix the failing test later). But it's hard to make it required only if Cassy's `resource_count > 0`. Do you think the `storage_base_uri` can be optional? If the URI is empty, [an incomplete IAM policy](https://github.com/scalar-labs/scalar-terraform/pull/103/files#diff-c344f8fb6da54549a7118914d9c30357R232) would be created, but it doesn't take effect.

ymorimo:
@feeblefakie @tei-k 
As we discussed, I updated the code to use the `aws_s3_bucket` data resource so that it can check the existence of the S3 bucket.
Then, the test need a bucket to succeed. I created a bucket named `scalar-terraform-test` and [set it in the tfvars](https://github.com/scalar-labs/scalar-terraform/pull/103/commits/3509177fe53cf0f0505bcccce12929d63515229e).

examples/aws/README.md
```diff
@@ -8,6 +8,7 @@ This example will deploy a simple Scalar DL environment in the Tokyo region with
 * Ansible 2.8
 * AWS CLI
 * ssh-agent with private key
+* An S3 bucket to store Cassandra backups with Cassy
```
feeblefakie, line 11:
```suggestion
* An S3 bucket URI to store Cassandra backups with Cassy
```

feeblefakie:
@ymorimo Oh can you fix the conflict ?
@tei-k Please take a final look !

modules/aws/cassandra/cassandra.tf
```diff
@@ -12,6 +12,7 @@ module "cassandra_cluster" {
   subnet_id                   = local.subnet_id
   associate_public_ip_address = false
   hostname_prefix             = "cassandra"
+  iam_instance_profile        = local.cassy.storage_base_uri != "" ? aws_iam_instance_profile.cassandra.name : ""
```
tei-k, line 15: (IMO) 
Since `aws_iam_instance_profile` is always created, I think don't need to make any condition. 🤔

modules/aws/cassandra/cassandra.tf
```diff
@@ -182,6 +183,79 @@ resource "null_resource" "volume_commitlog_local" {
...
+      "s3:GetObject",
+      "s3:DeleteObject",
+      "s3:PutObjectAcl",
+    ]
+    resources = [
+      "${data.aws_s3_bucket.cassy_storage[0].arn}/*"
+    ]
+  }
+}
+
```
tei-k, line 258: (IMO)

How about separating the iam role part into a separate file? (e.g. iam.tf)
It may be easier to maintain it if it is separated from the cassy variables because they are mixed up.

tei-k:
Looks good ! 👍 
Left some comments. Please take a look! 😅

modules/aws/cassandra/cassandra.tf
ymorimo: Thank you for the suggestion 👍. Created `iam.tf`.

modules/aws/cassandra/cassandra.tf
ymorimo: Good point. Updated!

docs/CassySetup.md
```diff
@@ -25,13 +27,11 @@ Cassy master tells each Cassandra node to upload backup files to Cloud Storages
...
 
     // the following variables need to be filled
-    scalar.cassy.server.storage_base_uri=
     scalar.cassy.server.srv_service_url=
     ```
 
     e.g.
     ```
-    scalar.cassy.server.storage_base_uri=s3://your-bucket-name
     scalar.cassy.server.srv_service_url=_scalardl._tcp.scalardl-service.internal.scalar-labs.com
```
feeblefakie, line 35: @ymorimo Please automatically fill this with whatever specified domain name in a next PR so that we can avoid manual operation for cassy setup.

PR #104, opened 2020-04-15
Add dev for azure

tei-k:
# Description
https://scalar-labs.atlassian.net/browse/DLT-4907


modules/azure/cassandra/locals.tf
```diff
@@ -41,6 +41,12 @@ locals {
   cassandra_base = {
     default = local.cassandra_default
 
+    dev = merge(local.cassandra_default,
+      {
+        resource_type = "Standard_E2s_v3"
```
feeblefakie, line 46: With the same way, I think Standard_B2ms is good enough for dev.

examples/azure/cassandra/example.tfvars
```diff
@@ -1,4 +1,4 @@
-base = "default" # bai, chiku, sho
+base = "default" # dev, bai, chiku, sho
```
feeblefakie, line 1:
```suggestion
base = "default" # dev, bai (default), chiku, sho
```

PR #106, opened 2020-04-16
Fix cassy module bugs

ymorimo:
This PR addresses two minor bugs of cassy module.

When applied after `null_resource.cassy_container[0]` is tainted:

1. `~/cassy/` directory is not updated correctly. It creates `~/cassy/provision/` instead.
2. The Cassy private key is read-only, so the `echo {key} > ~/.ssh/cassy.pem` command fails and the container cannot be started.

### DIFF
--- a/modules/universal/cassy/main.tf
+++ b/modules/universal/cassy/main.tf
@@ -72,7 +72,7 @@ resource "null_resource" "cassy_container" {
   provisioner "remote-exec" {
     inline = [
       "echo '${tls_private_key.cassy_private_key.private_key_pem}' > $HOME/.ssh/cassy.pem",
-      "chmod 400 $HOME/.ssh/cassy.pem",
+      "chmod 600 $HOME/.ssh/cassy.pem",
       "cd $HOME/cassy",
       "docker-compose up -d",
     ]
--- a/modules/universal/cassy/main.tf
+++ b/modules/universal/cassy/main.tf
@@ -65,7 +65,7 @@ resource "null_resource" "cassy_container" {
   }
 
   provisioner "file" {
-    source      = "${path.module}/provision"
+    source      = "${path.module}/provision/"
     destination = "$HOME/cassy"
   }
 


modules/universal/cassy/main.tf
```diff
@@ -65,14 +65,14 @@ resource "null_resource" "cassy_container" {
   }
 
   provisioner "file" {
-    source      = "${path.module}/provision"
+    source      = "${path.module}/provision/"
     destination = "$HOME/cassy"
```
tei-k, line 69: [Q] Why is the destination set to `cassy` instead of the `provision` ? (Sorry! It's not related to this PR)

e.g. Reaper
https://github.com/scalar-labs/scalar-terraform/blob/1d5c1314c3ce9a7b7fbd2140db05645983ccf9dc/modules/universal/reaper/main.tf#L62-L64

feeblefakie: Is the following OK for reaper ? Should it also be like this (having `/` at the tail) ?
```
source      = "${path.module}/provision" 
```

ymorimo: @tei-k I'm not sure but it's been like that from `scalardl-orchestration` tool.

modules/universal/cassy/main.tf
```diff
@@ -65,14 +65,14 @@ resource "null_resource" "cassy_container" {
...
   }
 
   provisioner "remote-exec" {
     inline = [
       "echo '${tls_private_key.cassy_private_key.private_key_pem}' > $HOME/.ssh/cassy.pem",
-      "chmod 400 $HOME/.ssh/cassy.pem",
+      "chmod 600 $HOME/.ssh/cassy.pem",
```
feeblefakie, line 75: Instead of changing the permission, should it delete `cassy.pem` first ?

modules/universal/cassy/main.tf
ymorimo: @feeblefakie It's ok for reaper because the destination is `$HOME`. The behavior is like rsync. https://www.terraform.io/docs/provisioners/file.html#directory-uploads

modules/universal/cassy/main.tf
ymorimo: `600` is a common mode for private keys because `ssh-keygen` generates a key with the mode. I don't think there is a special reason to make only the key file read-only.

ymorimo:
This PR introduced another bug that the initial deployment fails.
```
Error: Upload failed: scp: /home/centos/cassy: Not a directory
```

#107 fixes this problem.

PR #110, opened 2020-04-21
Configurable cassy version

tei-k:
# Description
https://scalar-labs.atlassian.net/browse/DLT-5754


examples/aws/cassandra/example.tfvars
```diff
@@ -20,6 +20,7 @@ cassandra = {
 }
 
 cassy = {
+  # version                   = "v1.0.0"
```
tei-k, line 23: ❓ `image_tag` is would be better?

ymorimo: I think `image_tag` would correspond to scalardl module.

tei-k: 🙆‍♂️I think so too. fixed in 22cdb06.

feeblefakie: Should this example be updated ?

tei-k: oh! Updated c62ab08

PR #117, opened 2020-05-01
Simplify the replacement steps

ymorimo:
https://scalar-labs.atlassian.net/browse/DLT-5552

This PR makes the provisioning in "universal" modules be triggered if the instances/VMs resources in aws/azure modules are tainted.

Following `cassandra` module's manner, a variable named `vm_ids` has been added to every `universal` module in addition to the `triggers` variable.

To replace a node, now we need to taint only the instance/VM resource and there's no need to taint the `null_resource` anymore.


ymorimo:
Following `cassandra` module's manner, I'm adding `vm_ids` variable in addition to the general `triggers` variable.

docs/NodeReplacement.md
```diff
@@ -29,100 +25,59 @@ module.network.module.bastion.module.bastion_provision.null_resource.ansible_pla
...
+module.scalardl.module.scalardl_blue.module.scalardl_cluster.aws_instance.this[0]
+# Azure
+module.scalardl.module.scalardl_blue.module.cluster.azurerm_virtual_machine.vm-linux[0]
+```
 
-    ```
-    module.scalardl.module.scalardl_blue.module.scalardl_provision.null_resource.scalardl_image[0]
-    ```
+* The index for `aws_instance.this` or `azurerm_virtual_machine.vm-linux` should be changed according to the node you want to replace.
+* If you are replacing a `green` node, change `blue` in the resource names with `green`.
```
feeblefakie, line 36:
```suggestion
* If you replace a `green` node, replace `blue` in the resource names with `green`.
```

PR #122, opened 2020-05-11
Add privileged port for envoy

tei-k:
# Description
https://scalar-labs.atlassian.net/browse/DLT-5955

# Done
- Add privileged port(`50052`).
- Delete `envoy-tls.yaml.j2` to using `envoy.yaml.j2`.
- Fix terratest to using default yaml.
- Fix yaml format.
- Fix resource name from `-` to `_`.


modules/universal/envoy/provision/envoy.yaml.j2
```diff
@@ -4,76 +4,91 @@ admin:
   address:
     socket_address: { address: 0.0.0.0, port_value: 9001 }
 
+{% set envoy_listeners = {'scalar-service': 50051, 'scalar-privileged': 50052} -%}
```
feeblefakie, line 7: Oh, it's nice !

ymorimo: @tei-k Is it okay here 50051 and 50052 are hard-coded while these values are passed to this module as variables?

tei-k: @ymorimo 

Good point!

These are for the ports in the container, so related to the port mapping in `docker-compose.yml`.
So I think it's ok for hard coating.

https://github.com/scalar-labs/scalar-terraform/blob/f76da26bcfbb462344165824789add77835f7c94/modules/universal/envoy/provision/docker-compose.yml#L7

ymorimo: @tei-k Ahh, I see. 👍 Thanks!

feeblefakie: I understand it now ! Thank you !

tei-k:
- scalar-service
Screen Shot 2020-05-14 at 12 12 47

- scalar-privileged
Screen Shot 2020-05-14 at 13 57 50

PR #123, opened 2020-05-13
Refactoring additional public keys

tei-k:
# Description

- example.tfvars
Comment out `additional_public_keys_path`
```
# additional_public_keys_path = "./additional_public_keys"
```
- file `./additional_public_keys` exists.

=> `additional_public_keys` is added.

# Done
- Fix default value of `additional_public_keys_path` to "".
- Fix readme with markdown format.


modules/universal/bastion/main.tf
```diff
@@ -18,8 +18,9 @@ resource "null_resource" "ansible_playbooks_copy" {
   provisioner "local-exec" {
     command = <<EOT
 # add addtional public keys to tmp file
-if [[ -s ${var.additional_public_keys_path} ]]; then 
-  cp ${var.additional_public_keys_path} ${module.ansible.local_playbook_path}/files/ssh/additional_public_keys; 
+:>| ${module.ansible.local_playbook_path}/files/ssh/additional_public_keys
```
feeblefakie, line 21: Just wondering. What is this `:>|` ?

ymorimo:
@tei-k I'm wondering if we should align table borders in README. Did you format the markdown with some tool? I think it should be left as terraform-docs generates because we don't always have the same formatter.

modules/universal/bastion/main.tf
tei-k: Just for clear contents of target file. (Can to use `rm -f` instead of `:>|`)

1. 最初に `var.additional_public_keys` を指定して実行すると`${module.ansible.local_playbook_path}/files/ssh/additional_public_keys` へファイルがコピーされる
2. 同じディレクトリ内で違う環境を作成したい場合に、`var.additional_public_keys` 指定を外しても1で作られたファイル`${module.ansible.local_playbook_path}/files/ssh/additional_public_keys`が残るため意図せず、追加keyとして登録される

tei-k:
> I'm wondering if we should align table borders in README. Did you format the markdown with some tool? I think it should be left as terraform-docs generates because we don't always have the same formatter.

@ymorimo 

Thanks! It's fixed by markdown extension of vscode. 💦

modules/universal/bastion/main.tf
ymorimo: I have just figured out that `: >| file` empties a file with the `noclobber` shell option disabled. There are a lot of ways to do it, but I think `echo -n > file` or `cat /dev/null > file` is okay because we don't expect the `noclobber` option is enabled here. (Personally I do `:> file` in my terminal because it's shortest, but it feels too magical to use in the script.)

penpyt: @tei-k nice catch on the multiple run, IMO `rm` command look better in a script

tei-k: >Personally I do :> file in my terminal because it's shortest

Me too. How about using `rm -f` to delete first.

feeblefakie: Oh, I didn't know that. 
But as @ymorimo says, I think it should be replaced with something else for readability.

tei-k: @penpyt 

Thanks! Fixed in 13f5e3b

ymorimo: @tei-k I thought it needed to be truncated without removing. `rm` is fine! 👍

modules/aws/network/output.tf
```diff
@@ -65,7 +65,7 @@ output "private_key_path" {
 }
 
 output "additional_public_keys_path" {
-  value       = abspath(pathexpand(var.additional_public_keys_path))
+  value       = trimspace(var.additional_public_keys_path) == "" ? "" : abspath(pathexpand(trimspace(var.additional_public_keys_path)))
```
ymorimo, line 68: With `trimspace()`, do we need to consider the possible spaces around the path? There is no other conditional expression like this. IMO if the user-provided path contains any extra space, it should simply fail.

tei-k: Good point! 👍 

It will succeed if var.additional_public_keys_path is "    ".
So I want to ouput "" in network output.

```
> abspath(pathexpand("   "))
/Users/tei/work/src/scalar/scalar-terraform/examples/aws/network/
```

ymorimo: The output of `abspath(pathexpand("   "))` is actually "`/Users/tei/work/src/scalar/scalar-terraform/examples/aws/network/   `" . It doesn't make any difference to that `abspath(pathexpand("foo"))` outputs "`/Users/.../network/foo`". "foo" also succeeds because it is ignored by the shell's `[[ -s ]]` condition. (Sorry but I said it should fail but it was wrong.) So I think there seems to be no reason to treat whitespaces specially.

tei-k: 処理的に影響ないのでどちらもよいかと思うが、
ファイルが存在するかshell内の処理とは関係なくoutputに current directoryが返されるのは微妙かなと思いspaceをケアした感じです〜 ( `local_file` resourceとかでリファクタリングするまでもなかったので)


- var="foo", output="/Users/.../network/foo" => OK
- var="   ", output="/Users/.../network/" => NG

tei-k: @ymorimo 

Fixed in (78f98f1)
- Recoverd output.tf of moudle.
- Deleted from output.tf of example

feeblefakie:
@ymorimo Please take a look again and approve if it is OK.

modules/aws/network/output.tf
ymorimo: Thanks!

PR #125, opened 2020-05-14
Upgrade Cassy to 1.2.0 and add Azure managed identity

ymorimo:
https://scalar-labs.atlassian.net/browse/DLT-5554

This PR upgrades Cassy to version 1.2.0 and assigns a managed identity to Cassandra and Cassy VMs which allows uploading backups to the blob storage.


modules/azure/cassandra/identity.tf
```diff
@@ -0,0 +1,29 @@
+data "azurerm_subscription" "subscription" {}
+
+data "azurerm_role_definition" "contributor" {
+  name = "Storage Blob Data Contributor"
+}
+
+data "azurerm_storage_account" "cassy-storage-account" {
```
tei-k, line 7: Let's use underscore like `cassy_storage_account`.
https://www.terraform.io/docs/extend/best-practices/naming.html#resource-names

test/modules/azuredeploy/cassandra/terraform.tfvars
```diff
@@ -3,5 +3,6 @@ cassandra = {
 }
 
 cassy = {
-  resource_root_volume_size = 16
+  # Set the count to zero because it needs a storage account
+  resource_count = 0
```
tei-k, line 7: It's not a CASSY's resource creation test. 🤔

feeblefakie:
@ymorimo BTW, Boney was able to take backup and restore without manual configuration.
https://scalar-labs.atlassian.net/browse/MT-109

modules/azure/cassandra/identity.tf
ymorimo: Thanks! Fixed. b56dd30

test/modules/azuredeploy/cassandra/terraform.tfvars
ymorimo: `terraform plan` fails because it tries to read an actual storage account by the `data` resource. https://github.com/scalar-labs/scalar-terraform/runs/692368211
I set the `resource_count` to zero just to skip this part.

In AWS we can keep an S3 bucket just for the test because the bucket is not related to any VPC. https://github.com/scalar-labs/scalar-terraform/blob/v1.2.0/test/modules/awsdeploy/cassandra/terraform.tfvars#L9
But in Azure, we can't prepare a storage account in advance because it must be created in a resource group.

tei-k: I understand what you mean. Just worried can't test for creation cassy in Azure.

I'm not sure, But can we use flag or create about storage count resoruce in terratest

modules/azure/cassandra/locals.tf
```diff
@@ -86,12 +86,15 @@ locals {
...
 locals {
   cassy_default = {
-    image_tag                     = "v1.0.0"
+    image_tag                     = "1.2.0"
     resource_type                 = "Standard_B2s"
     resource_count                = 1
     resource_root_volume_size     = 64
     enable_tdagent                = true
     enable_accelerated_networking = false
+    storage_account_name          = ""
```
feeblefakie, line 95: I think it would be more user friendly if `storage_account_name` is extracted from `storage_base_uri` because it is duplicate, but is it not easy ?

provision/ansible/playbooks/roles/cassandra/files/cassy/build/install/cassy/bin/cassy-backup
```diff
@@ -1,5 +1,21 @@
 #!/usr/bin/env sh
```
feeblefakie, line 1: It is not scope of this PR, but do you have any better idea for installing the commands instead of copying all the files. > @ymorimo @tei-k

modules/azure/cassandra/locals.tf
ymorimo: Yes, it's kind of a duplication to have `storage_account_name` separately. I chose this because I'm not sure if we can assume that the `storage_base_uri` is always formed like `<storage account name>` + `blob.core.windows.net`.
Technically it is possible to get the storage account name from the URI by assuming the first part of the domain is the name.

provision/ansible/playbooks/roles/cassandra/files/cassy/build/install/cassy/bin/cassy-backup
ymorimo: There seems no need to copy whole the `build` directory. Only `build/install/cassy` that is created by `./gradlew installDist` can do the job on cassandra hosts.

```console
$ ls /opt/cassy/build/install/cassy/
bin  cassy.jar  lib
```

feeblefakie: Yeah, that can be done pretty easily. Let's start from there !

feeblefakie: https://scalar-labs.atlassian.net/browse/DLT-6083

examples/azure/README.md
```diff
@@ -55,6 +55,30 @@ $ terraform apply -var-file example.tfvars
 
 ### Create Cassandra resources
 
+To use Cassy to backup Cassandra data (enabled by default), you need to create a storage account before creating resources with `scalar-terraform`.
```
feeblefakie, line 58:
```suggestion
Before creating Cassandra resources with `terraform apply`, you probably need to configure for Cassy to manage backups of Cassandra data. 

The first thing you need to do for Cassy is create a storage account in the same resource group as the network resource created in the previous section and create a blob type container in the storage account.

Then, update `example.tfvars` with the container URL as follows.
```

examples/azure/README.md
```diff
@@ -55,6 +55,30 @@ $ terraform apply -var-file example.tfvars
...
+Note that it must be in the resource group created by the network module in the previous section.
```
feeblefakie, line 59: Please delete this if the above suggestion is committed.

examples/azure/README.md
```diff
@@ -55,6 +55,30 @@ $ terraform apply -var-file example.tfvars
...
+
+Then, create a blob type container in the storage account, and update `example.tfvars` as follows.
```
feeblefakie, line 61: Please delete this if the above suggestion is committed.

examples/azure/README.md
```diff
@@ -55,6 +55,30 @@ $ terraform apply -var-file example.tfvars
...
+
+```
+cassy = {
+  storage_base_uri     = "https://yourstorageaccountname.blob.core.windows.net/your-container-name"
+  storage_type         = "azure_blob"
+}
+```
+
+You can also turn off the use of Cassy by setting its `resource_count` to zero.
```
feeblefakie, line 70:
```suggestion
If you don't need Cassy, you can disable it by setting its `resource_count` to zero.
```

examples/azure/README.md
```diff
@@ -55,6 +55,30 @@ $ terraform apply -var-file example.tfvars
...
+
+```
+cassy = {
+  resource_count = 0
+}
+```
+
+For more information on Cassy, please refer to [CassySetup](../../docs/CassySetup.md).
+
+Then run the terraform command:
```
feeblefakie, line 80:
```suggestion
Now it's ready to run the terraform commands:
```

PR #126, opened 2020-05-15
Upgrade Envoy version to 1.14.1

ymorimo:
https://scalar-labs.atlassian.net/browse/DLT-5334

https://www.envoyproxy.io/docs/envoy/latest/version_history/v1.14.0#deprecated


modules/universal/envoy/provision/envoy.yaml.j2
```diff
@@ -15,16 +15,16 @@ static_resources:
           port_value: {{ value|e }}
       filter_chains:
         - filters:
-          - name: envoy.http_connection_manager
+          - name: envoy.filters.network.http_connection_manager
             config:
```
tei-k, line 19: [Q]
`typed_config:` used in `transport_socket` session, Is it better to use `typed_config` instead of `config` ?

https://github.com/envoyproxy/envoy/blob/32b8445b8fac9e46f67c2f2ce884716b7bc9dde8/configs/envoy_front_proxy_v2.template.yaml#L38
```
typed_config:
  "@type": type.googleapis.com/envoy.config.filter.network.http_connection_manager.v2.HttpConnectionManager
```

ymorimo: @tei-k Thank you for pointing out. 👍 I didn't notice that. It would be better to follow the example but I'm not sure what is `typed_config`. Do you know if `config` is being deprecated?

feeblefakie: @tei-k @ymorimo Any further discussion is needed ? Let's wrap it up.

feeblefakie:
@ymorimo There is a conflict also. Please fix !

modules/universal/envoy/provision/envoy.yaml.j2
tei-k: @ymorimo 

Sorry! Sorry for the late response.
I'm not sure, But `config` seems to be deprecated.

https://www.envoyproxy.io/docs/envoy/latest/version_history/v1.14.0#deprecated
```
config: use type URL to select an extension whenever the config type URL (or its previous versions) uniquely identify a typed extension, see extension configuration
```
https://www.envoyproxy.io/docs/envoy/latest/configuration/overview/extension#config-overview-extension-configuration

ymorimo: @tei-k Thanks! I'll update the file to use `typed_config`.

ymorimo: Done in ec7190e

feeblefakie:
@tei-k It should go to 1.3 ? or still going to 1.2.x ?

tei-k:
>It should go to 1.3 ? or still going to 1.2.x ?

@feeblefakie 

It's true, I think it's better to upgrade a little at a time.
It would be good if version 1.14 is enough to test.

PR #134, opened 2020-05-21
Refactoring resource name

tei-k:
Fix `-` to `_` in resource name.


feeblefakie:
@tei-k Just a question. It's only dns resource name renaming so it can go to 1.2.x ? 
It might trigger other resource recreation except dns ? (so it's should be in 1.3.0 ?)

tei-k:
>Just a question. It's only dns resource name renaming so it can go to 1.2.x ?
It might trigger other resource recreation except dns ? (so it's should be in 1.3.0 ?)

It's ok for 1.2.x, But need to apply again when first apply was failed.

PR #135, opened 2020-05-25
Fix to use setup-terraform for aws

tei-k:
# Description
https://scalar-labs.atlassian.net/browse/DLT-6037

refs:
https://github.com/hashicorp/setup-terraform/

### DIFF
--- a/.github/workflows/aws_ca.yml
+++ b/.github/workflows/aws_ca.yml
@@ -45,7 +45,7 @@ jobs:
           mv locals.tf.dummy locals.tf
 
       - name: Terraform Format
-        run: terraform fmt -check
+        run: terraform fmt -check -recursive
         working-directory: ./modules/aws/ca
 
       - name: Terraform Init
--- a/.github/workflows/aws_ca.yml
+++ b/.github/workflows/aws_ca.yml
@@ -1,65 +1,59 @@
-name: 'Terraform test for CA module of AWS'
+name: Terraform test for CA module of AWS
+
 on:
   pull_request:
     branches:
       - master
     paths:
-      - 'modules/aws/ca/**'
-      - 'modules/universal/ca/**'
-      - 'test/modules/awsdeploy/ca/**'
-      - 'provision/ansible/**'
+      - modules/aws/ca/**
+      - modules/universal/ca/**
+      - test/modules/awsdeploy/ca/**
+      - provision/ansible/**
   push:
     branches:
       - master
     paths:
-      - 'modules/aws/ca/**'
-      - 'modules/universal/ca/**'
-      - 'test/modules/awsdeploy/ca/**'
-      - 'provision/ansible/**'
+      - modules/aws/ca/**
+      - modules/universal/ca/**
+      - test/modules/awsdeploy/ca/**
+      - provision/ansible/**
+
 jobs:
   terraform:
-    name: 'Terraform'
+    name: Terraform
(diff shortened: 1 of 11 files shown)


tei-k:
@penpyt 

We're both fixing the workflow as well, So please take a look at it when you have time.

.github/workflows/aws_ca.yml
```diff
@@ -1,65 +1,58 @@
...
     runs-on: ubuntu-latest
     env:
       AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID }}
       AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
-      AWS_DEFAULT_REGION: ${{ secrets.AWS_DEFAULT_REGION }}
-      TF_VERSION: 0.12.19
-      TF_ACTIONS_COMMENT: false
-      TF_TEST_WORKING_DIR: ./test/modules/awsdeploy/ca
+      TF_VERSION: 0.12.25
+    defaults:
```
penpyt, line 29: it make more sense to put the `defaults` at the root level than job, what do think ?
https://help.github.com/en/actions/reference/workflow-syntax-for-github-actions#defaultsrun

.github/workflows/aws_cassandra.yml
```diff
@@ -1,69 +1,62 @@
...
-      - 'modules/universal/cassy/**'
-      - 'modules/universal/reaper/**'
-      - 'test/modules/awsdeploy/cassandra/**'
-      - 'provision/**'
+      - modules/aws/cassandra/**
+      - modules/universal/cassandra/**
+      - modules/universal/cassy/**
+      - modules/universal/reaper/**
+      - test/modules/awsdeploy/cassandra/**
+      - provision/**"
```
penpyt, line 13: the double quotes here create an error, no ? ;)

.github/workflows/aws_network.yml
```diff
@@ -1,62 +1,55 @@
...
-        with:
-          tf_actions_subcommand: 'plan'
-          tf_actions_working_dir: ${{ env.TF_TEST_WORKING_DIR }}
-          tf_actions_version: ${{ env.TF_VERSION }}
-          tf_actions_comment: ${{ env.TF_ACTIONS_COMMENT }}
+          terraform_version: ${{ env.TF_VERSION }}
+
+      - name: Terraform Format
+        run: terraform fmt -check
+        working-directory: ./modules/aws/network
```
penpyt, line 46: no need to put working-directory as it is defined in the defaults

.github/workflows/aws_ca.yml
tei-k: Both are ok.
Since there is only one job in one workflow, so I defined the default value in job level.

.github/workflows/aws_cassandra.yml
tei-k: 😱 delete misstake. fixed in 1f89641.

.github/workflows/aws_network.yml
tei-k: Thanks!

Default working-directory is `./test/modules/awsdeploy/network`. I need check module directory as `./modules/aws/network`.

penpyt: ok sorry my mistake

PR #138, opened 2020-05-27
Fix log volume re-attachment issue on monitor node

ymorimo:
https://scalar-labs.atlassian.net/browse/DLT-5545

1. Make the external log volume independent of the instance

    The log volume was being deleted when the instance is removed because the availability zone was unknown. This part is changed in the same way as C* nodes.

```
  # module.monitor.aws_ebs_volume.monitor_log_volume[0] must be replaced
-/+ resource "aws_ebs_volume" "monitor_log_volume" {
      ~ arn               = "arn:aws:ec2:us-west-1:825543958249:volume/vol-0259b0b418eadfe33" -> (known after apply)
      ~ availability_zone = "us-west-1b" -> (known after apply) # forces replacement
      ~ encrypted         = false -> (known after apply)
      ~ id                = "vol-0259b0b418eadfe33" -> (known after apply)
      ~ iops              = 0 -> (known after apply)
      + kms_key_id        = (known after apply)
        size              = 500
      + snapshot_id       = (known after apply)
        tags              = {
            "Name"      = "ymtest-i7eqxem Monitor Cluster-1"
            "Network"   = "ymtest-i7eqxem"
            "Owner"     = "yusuke"
            "Terraform" = "true"
        }
        type              = "sc1"
    }
```

2. Update drive_mount script so that it can mount an existing log volume correctly

   The log volume re-attachment was not working because `drive_mount.sh` was not able to handle an external volume that is already formatted. Now it can add an `/etc/fstab` entry correctly even if the volume is not new.

### DIFF
--- a/modules/azure/monitor/main.tf
+++ b/modules/azure/monitor/main.tf
@@ -33,7 +33,6 @@ resource "azurerm_managed_disk" "monitor_log_volume" {
   storage_account_type = local.monitor.log_volume_type
   create_option        = "Empty"
   disk_size_gb         = local.monitor.log_volume_size
-  depends_on           = [module.monitor_cluster]
 }
 
 resource "azurerm_virtual_machine_data_disk_attachment" "monitor_log_volume_attachment" {
--- a/provision/ansible/playbooks/roles/td-agent_aggregation/files/drive_mount.sh
+++ b/provision/ansible/playbooks/roles/td-agent_aggregation/files/drive_mount.sh
@@ -1,5 +1,49 @@
 #!/bin/bash
 
+# The log volume is already mounted by fstab entry on startup
+if mountpoint -q /log; then
+  # Do only chown
+  chown td-agent:td-agent /log
+  exit 0
+fi
+
+# Use root volume if LOG_STORE is not set
+if [[ -z "$LOG_STORE" ]]; then
+  # Create /log directory at the root
+  echo "LOG_STORE is not set: Using Root Volume"
+  mkdir -p /log
+  chown td-agent:td-agent /log
+  exit 0
+fi
+
+mount_log_volume() {
+  name=$1
+  uuid=$2
+
+  echo "Using $NAME as a log volume"
+
+  if [[ -z $uuid ]]; then
+    # The volume is not formatted
+    echo "Running mkfs on $NAME"
+    if mkfs -t xfs $name; then
+      uuid=$(blkid $name -s UUID -o value)
+    else
+      echo "mkfs failed"
+      exit 1
+    fi
+  fi
+
+  if ! grep -q "^UUID=$uuid" /etc/fstab; then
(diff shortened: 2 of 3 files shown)


modules/azure/monitor/main.tf
```diff
@@ -33,7 +33,6 @@ resource "azurerm_managed_disk" "monitor_log_volume" {
   storage_account_type = local.monitor.log_volume_type
   create_option        = "Empty"
   disk_size_gb         = local.monitor.log_volume_size
-  depends_on           = [module.monitor_cluster]
```
tei-k, line 36: Do we need this?
https://github.com/scalar-labs/scalar-terraform/blob/fa34000ba3ff9addd9d579ec12ce7e6a9e9ac98b/modules/azure/cassandra/cassandra.tf#L26

provision/ansible/playbooks/roles/td-agent_aggregation/files/drive_mount.sh
```diff
@@ -1,51 +1,70 @@
...
-'
+# The log volume is already mounted by fstab entry on startup
+if mountpoint -q /log; then
+  # Do only chown
+  chown td-agent:td-agent /log
+  exit 0
+fi
+
+# Use root volume if LOG_STORE is not set
+if [[ -z "$LOG_STORE" ]]; then
```
feeblefakie, line 11: I don't think we should use OS drive for this logging in any case for production so I think it should simply fail.
What do you think ?

ymorimo: I agree! I added this part because the option `enable_log_volume` can be `false`. So this option should be removed together.

ymorimo: Using the root volume when there is no log volume is the same behavior as the previous drive_mount script.
The `enable_log_volume` flag was intended to support cases where we configure td-agent to store logs on locations other than a local disk, such as S3. But we can remove the flag for now since custom configuration of td-agent is not supported yet.

feeblefakie: Sounds good. Let's do that !

ymorimo: Okay 👍 I'd like to create another story about that since it is a different topic than the re-attachment issue.

feeblefakie: Yes, that would be great.

ymorimo: I'd like to keep this `if` section for now so that the script works as before.

PR #139, opened 2020-05-27
Fix mount issue in monitor module

tei-k:
# Description
https://scalar-labs.atlassian.net/browse/DLT-6088

# Done
- Add triggers for universal monitor to control the order of creation
- Add triggers for aws as Azure

# Did't done
- Change operation for `enable_log_volume`
`true -> false` or `false -> true`
※ this may be related to issue of re-attachment.


feeblefakie:
Looks good but it should be aligned with #138 so let's talk about them together tomorrow.

PR #141, opened 2020-06-02
Ignore use_managed_identity flag if cassy's resource_count is zero

ymorimo:
https://scalar-labs.atlassian.net/browse/DLT-6116

Resolves an issue that terraform fails with an index error when `resource_count` of Cassy is zero and `use_managed_identity` is `true` (default). Now it doesn't use `use_managed_identity` if `resource_count` is zero.

scalar-labs/scalar#465

### DIFF
--- a/modules/azure/cassandra/identity.tf
+++ b/modules/azure/cassandra/identity.tf
@@ -13,7 +13,7 @@ data "azurerm_storage_account" "cassy_storage_account" {
 }
 
 resource "azurerm_role_assignment" "cassy" {
-  count = local.cassy.use_managed_identity ? local.cassy.resource_count : 0
+  count = local.cassy.use_managed_identity && length(data.azurerm_storage_account.cassy_storage_account) > 0 ? local.cassy.resource_count : 0
 
   scope              = data.azurerm_storage_account.cassy_storage_account[0].id
   role_definition_id = "${data.azurerm_subscription.subscription.id}${data.azurerm_role_definition.contributor.id}"
@@ -21,7 +21,7 @@ resource "azurerm_role_assignment" "cassy" {
 }
 
 resource "azurerm_role_assignment" "cassandra" {
-  count = local.cassy.use_managed_identity ? local.cassandra.resource_count : 0
+  count = local.cassy.use_managed_identity && length(data.azurerm_storage_account.cassy_storage_account) > 0 ? local.cassandra.resource_count : 0
 
   scope              = data.azurerm_storage_account.cassy_storage_account[0].id
   role_definition_id = "${data.azurerm_subscription.subscription.id}${data.azurerm_role_definition.contributor.id}"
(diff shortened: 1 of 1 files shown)


modules/azure/cassandra/identity.tf
```diff
@@ -13,15 +13,15 @@ data "azurerm_storage_account" "cassy_storage_account" {
 }
 
 resource "azurerm_role_assignment" "cassy" {
-  count = local.cassy.use_managed_identity ? local.cassy.resource_count : 0
+  count = local.cassy.use_managed_identity && length(data.azurerm_storage_account.cassy_storage_account) > 0 ? local.cassy.resource_count : 0
```
ymorimo, line 16: This change is not really necessary, but it is the same as the condition for `resource "azurerm_role_assignment" "cassandra"` below, and it makes sense to check the length of an array before its element is accessed.

feeblefakie: Sorry I'm confused now. 
The previous code looks fine (?) since count is always 0 if cassy.resouce_count is 0, but why it was not ?
https://github.com/scalar-labs/scalar/pull/465
(it says it needs to set both cassy.resouce_count = 0 and use_managed_identity = false.

modules/azure/cassandra/identity.tf
```diff
@@ -13,15 +13,15 @@ data "azurerm_storage_account" "cassy_storage_account" {
...
 
   scope              = data.azurerm_storage_account.cassy_storage_account[0].id
   role_definition_id = "${data.azurerm_subscription.subscription.id}${data.azurerm_role_definition.contributor.id}"
   principal_id       = module.cassy_cluster.vm_identities[count.index][0]["principal_id"]
 }
 
 resource "azurerm_role_assignment" "cassandra" {
-  count = local.cassy.use_managed_identity ? local.cassandra.resource_count : 0
+  count = local.cassy.use_managed_identity && length(data.azurerm_storage_account.cassy_storage_account) > 0 ? local.cassandra.resource_count : 0
```
tei-k, line 24: ditto.

tei-k, line 24: ditto.

```suggestion
  count = length(data.azurerm_storage_account.cassy_storage_account) > 0 ? local.cassandra.resource_count : 0
```

modules/azure/cassandra/identity.tf
ymorimo: Yes, the problem was not this resource, but this block below. https://github.com/scalar-labs/scalar-terraform/pull/141/files#diff-aad39056f3291377f4c7510643d43ff0L24
I changed this line just to unify the two conditions. But as Tei san suggested, they could be simpler.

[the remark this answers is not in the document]
ymorimo: Right! Applied. Thanks 👍

PR #145, opened 2020-06-16
Able to select what to monitor

tei-k:
# Description
https://scalar-labs.atlassian.net/browse/DLT-6271

# Done
- Add `target_monitoring` vars for select target.
- Able to select `cassandra` and `scalardl`
`cassandra` include `cassandra`, `cassy`, `reaper`
`scalardl` include `scalardl-blue`, `scalardl-green`, `envoy`


examples/aws/monitor/vars.tf
```diff
@@ -12,6 +12,14 @@ variable "monitor" {
   default = {}
 }
 
+variable "target_monitoring" {
```
feeblefakie, line 15:
```suggestion
variable "targets" {
```
It's a variable in monitor module so `monitoring` is probably redundant.

examples/aws/monitor/vars.tf
```diff
@@ -12,6 +12,14 @@ variable "monitor" {
...
+  type = list(string)
+  default = [
+    "cassandra",
+    "scalardl",
```
feeblefakie, line 19: I guess the answer is `no` but let me ask a question. We don't really monitor the resources created by `network` module at the moment ?

modules/universal/monitor/locals.tf
```diff
@@ -0,0 +1,9 @@
+locals {
+  scalardl_targets  = contains(var.target_monitoring, "scalardl") ? ["scalardl-blue", "scalardl-green", "envoy"] : []
+  cassandra_targets = contains(var.target_monitoring, "cassandra") ? ["cassandra", "cassy", "reaper"] : []
+  general_targets   = ["grafana", "alertmanager", "prometheus", "nginx"]
```
feeblefakie, line 4: `general` is probably too blur. Seems more reasonable to use the module name since they are grouped by the modules.
BTW, I'm not sure if we need `nginx`. Do we still use it ?
So, should it be something like this ?
```
  monitor_targets   = ["grafana", "alertmanager", "prometheus"]
```

modules/universal/monitor/locals.tf
```diff
@@ -0,0 +1,9 @@
...
+
+  service_targets   = setunion(local.scalardl_targets, local.cassandra_targets)
+  node_targets      = setunion(local.service_targets, ["bastion", "monitor"])
+  cadvisor_targets  = setunion(local.service_targets, ["monitor"])
```
feeblefakie, line 8: It seems including `cassandra` but is it OK ?

examples/azure/monitor/example.tfvars
```diff
@@ -12,6 +12,12 @@ monitor = {
   # set_public_access             = "false"
   # remote_port                   = 9090
   # enable_accelerated_networking = "false"
+  # target_monitoring             = "cassandra,scalardl"
```
ymorimo, line 15: This isn't needed since there is another `target_monitoring` below?

modules/universal/monitor/provision/prometheus/general_alert.rules.yml.j2
```diff
@@ -0,0 +1,59 @@
...
+        labels:
+          severity: critical
+        annotations:
+          summary: "The td-agent process is down"
+          description: "The td-agent process on {{ '{{' }} $labels.instance {{ '}}' }} is down"
+
+  - name: Docker Container Alerts
+    rules:
+    {% for s in general_list + cassandra_list %}
+      - alert: {{ s }}ContainerDown
```
ymorimo, line 52: The alert name would be the same as before if the alert name is capitalized.

```suggestion
      - alert: {{ s|capitalize }}ContainerDown
```

modules/universal/monitor/provision/prometheus/general_alert.rules.yml.j2
```diff
@@ -0,0 +1,59 @@
...
+        expr: time() - container_last_seen{container_label_com_docker_compose_service="{{ s }}"} > 15
+        labels:
+          severity: warning
+        annotations:
+          summary: "{{ s }} container is down"
```
ymorimo, line 57: This summary was capitalized as well.

```suggestion
          summary: "{{ s|capitalize }} container is down"
```

modules/universal/monitor/vars.tf
```diff
@@ -60,3 +60,8 @@ variable "enable_tdagent" {
 variable "internal_domain" {
   description = "Internal domain"
 }
+
+variable "target_monitoring" {
+  default     = []
+  description = "Target monitoring"
```
ymorimo, line 66: This should be more descriptive. "A list of targets to be monitored" ?

examples/aws/monitor/vars.tf
tei-k: Fixed in 55b8e78

examples/aws/monitor/vars.tf
tei-k: No. The resources of `bastion` and `monitor` are always monitored.

modules/universal/monitor/locals.tf
tei-k: Fixed to `monitor_targets`. (I referenced the file name `general_alert_rules.yml`)

About nginx, Because the nginx container is used as a proxy.
client -> 8000 -> port forwarding 80 -> monitor (nginx) -> (grafana,prometheus, alertmanager)

So I think should consider it in another PR.

modules/universal/monitor/locals.tf
tei-k: Good point!

It doesn't affect me because `cassandra` is not registered with DNS, 
but I'll see if there's a better way.

examples/azure/monitor/example.tfvars
tei-k: Oh! Deleted. 💦

modules/universal/monitor/provision/prometheus/general_alert.rules.yml.j2
tei-k: Good point!

Fixed in c0342b4.

modules/universal/monitor/provision/prometheus/general_alert.rules.yml.j2
tei-k: Ditto.

modules/universal/monitor/vars.tf
tei-k: Looks good, But I'm not sure. call @feeblefakie

modules/universal/monitor/locals.tf
feeblefakie: This is not enough ?
```
  cadvisor_targets  = setunion(local.scalardl_targets, ["monitor"])
```

feeblefakie: @tei-k How about this ?

tei-k: Fixed in a4a9fba.

modules/universal/monitor/locals.tf
```diff
@@ -0,0 +1,10 @@
+locals {
+  scalardl_targets      = contains(var.targets, "scalardl") ? ["scalardl-blue", "scalardl-green", "envoy"] : []
+  cassandra_gen_targets = contains(var.targets, "cassandra") ? ["cassandra"] : []
+  cassandra_ops_targets = contains(var.targets, "cassandra") ? ["cassy", "reaper"] : []
```
ymorimo, line 4: What does `gen` and `ops` mean?
Since these two vars are never used outside this block, they could be made inline where they are used.

tei-k: Thansk! Fixed in 3ee3c7f.

PR #146, opened 2020-06-16
Enable Cassandra authentication and authorization

ymorimo:
https://scalar-labs.atlassian.net/browse/DLT-6244

* Updates cassandra.yaml to use the following
 
   ```
    authenticator: PasswordAuthenticator
    authorizer: CassandraAuthorizer
    ```

   `CassandraAuthorizer` is needed to run `GRANT` permissions to a specific user.

* Updates the default username and password for `scalardl` and `reaper` modules to authenticate to the C* cluster. The initial superuser `cassandra` is used by default.

We decided not to run any CQL commands such as `ALTER ROLE` and `CREATE ROLE` since they need a running Cassandra cluster.


ymorimo:
One of the tests crashes💥 for some reason...
@tei-k Can you find any reason?

PR #147, opened 2020-06-17
Disable debug.log of Cassandra

ymorimo:
https://scalar-labs.atlassian.net/browse/DLT-6294

Stops C* from writing debug.log by updating the logback.xml installed by the cassandra yum package, following the instructions in the file.

```
 <!--
In order to disable debug.log, comment-out the ASYNCDEBUGLOG
appender reference in the root level section below.
-->
```

The changes of llogback.xml are as follows.

```diff
   <root level="INFO">
     <appender-ref ref="SYSTEMLOG" />
     <appender-ref ref="STDOUT" />
-    <appender-ref ref="ASYNCDEBUGLOG" /> <!-- Comment this line to disable debug.log -->
+    <!-- <appender-ref ref="ASYNCDEBUGLOG" /> --> <!-- Comment this line to disable debug.log -->
     <!--
     <appender-ref ref="LogbackMetrics" />
     -->
```

### DIFF
--- a/provision/ansible/playbooks/roles/cassandra/tasks/main.yml
+++ b/provision/ansible/playbooks/roles/cassandra/tasks/main.yml
@@ -133,6 +133,12 @@
     src: cassandra-rackdc.properties.j2
     dest: /etc/cassandra/conf/cassandra-rackdc.properties
 
+- name: Disable debug.log
+  replace:
+    path: /etc/cassandra/conf/logback.xml
+    regexp: '(<appender-ref ref="ASYNCDEBUGLOG"\s*/>)'
+    replace: '<!-- \1 -->'
+
 - name: Delete cassandra-topology.properties
   file:
     path: /etc/cassandra/conf/cassandra-topology.properties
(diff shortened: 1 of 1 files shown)


feeblefakie:
@ymorimo This CI fails but #145 CI doesn't fail ... Hmm

ymorimo:
`terraform validate` in this workflow seems to keep crashing. It fails also in #146.
https://github.com/scalar-labs/scalar-terraform/blob/master/.github/workflows/aws_scalardl.yml
This is run because the change in this PR matches `provision/**`. (#145 doesn't trigger it.)

provision/ansible/playbooks/roles/cassandra/tasks/main.yml
```diff
@@ -133,6 +133,12 @@
     src: cassandra-rackdc.properties.j2
     dest: /etc/cassandra/conf/cassandra-rackdc.properties
 
+- name: Disable debug.log
+  replace:
+    path: /etc/cassandra/conf/logback.xml
+    regexp: '(<appender-ref ref="ASYNCDEBUGLOG"\s*/>)'
```
tei-k, line 139: Maybe better. 🤔 
(I'm a little concerned in the case of many runs.)
```
regexp: '(^\s*<appender-ref ref="ASYNCDEBUGLOG"\s*/>)'
```

ymorimo: Good catch. Updated in d810e76 (changing `(^\s*` to `^\s*(`) so that it is safe if this replacement happens to run multiple times.

PR #150, opened 2020-06-24
Fix td-agent.conf on bastion

ymorimo:
https://scalar-labs.atlassian.net/browse/DLT-6466

td-agent.conf for Cassandra hosts is installed on the bastion host.
Only the difference is that td-agent for Cassandra hosts looks at `/var/log/cassandra/system.log`.


feeblefakie:
@ymorimo Is this the reason why you said it wasn't working ?

ymorimo:
> @ymorimo Is this the reason why you said it wasn't working ?

@feeblefakie 
No, that was my misunderstanding of the warning message. 😑
https://scalar-labs.atlassian.net/browse/DLT-6467

PR #154, opened 2020-06-29
Fix default to use private lb for envoy in azure

tei-k:
# Description
https://scalar-labs.atlassian.net/browse/DLT-6543

# Done
- Fix default to use private lb for envoy in azure.
- Create public ip only when `nlb_internal` is `true`.
- Add `envoy-lb` dns for private dns.

# Confirm
- When `nlb_internal` = `false`
```
Outputs:

envoy_dns = envoy-tei-envoy-4u1raoq.japaneast.cloudapp.azure.com
```
- When `nlb_internal` = `true`
```
Outputs:

envoy_dns = envoy-lb.internal.scalar-labs.com
```
=>
```
[centos@bastion-1 ~]$ dig +short envoy-lb.internal.scalar-labs.com
10.42.1.7
```

### DIFF
--- a/modules/azure/scalardl/envoy.tf
+++ b/modules/azure/scalardl/envoy.tf
@@ -62,7 +62,7 @@ resource "azurerm_network_security_rule" "envoy_privileged_nsg" {
 }
 
 resource "azurerm_public_ip" "envoy_public_ip" {
-  count      = local.envoy.enable_nlb ? 1 : 0
+  count      = local.envoy.enable_nlb && ! local.envoy.nlb_internal ? 1 : 0
   depends_on = [null_resource.envoy_wait_for]
 
   name                = "PublicIPForEnvoy"
@@ -82,7 +82,7 @@ resource "azurerm_lb" "envoy_lb" {
   frontend_ip_configuration {
     name                          = "EnvoyLBAddress"
     public_ip_address_id          = local.envoy.nlb_internal ? "" : azurerm_public_ip.envoy_public_ip.*.id[count.index]
-    subnet_id                     = local.envoy_nlb_subnet_id
+    subnet_id                     = local.envoy.nlb_internal ? local.envoy_nlb_subnet_id : ""
     private_ip_address_allocation = "dynamic"
   }
 }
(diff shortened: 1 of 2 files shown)


modules/azure/scalardl/output.tf
```diff
@@ -39,7 +39,7 @@ output "scalardl_replication_factor" {
 }
 
 output "envoy_dns" {
-  value       = local.envoy.resource_count > 0 ? azurerm_public_ip.envoy_public_ip.*.fqdn : []
+  value       = local.envoy.enable_nlb ? (local.envoy.nlb_internal ? ["envoy-lb.${local.internal_domain}"] : azurerm_public_ip.envoy_public_ip.*.fqdn) : []
```
tei-k, line 42: 📝  `fqdn` does not worked for me.
https://www.terraform.io/docs/providers/azurerm/r/private_dns_a_record.html#fqdn
```
Error: Unsupported attribute
  on ../../../modules/azure/scalardl/output.tf line 42, in output "envoy_dns":
  42:   value       = local.envoy.enable_nlb ? (local.envoy.nlb_internal ? [azurerm_private_dns_a_record.envoy_lb[0].fqdn] : azurerm_public_ip.envoy_public_ip.*.fqdn) : []
This object has no argument, nested block, or exported attribute named "fqdn".
```

modules/azure/scalardl/envoy.tf
```diff
@@ -62,7 +62,7 @@ resource "azurerm_network_security_rule" "envoy_privileged_nsg" {
 }
 
 resource "azurerm_public_ip" "envoy_public_ip" {
-  count      = local.envoy.resource_count > 0 ? 1 : 0
+  count      = local.envoy.enable_nlb && ! local.envoy.nlb_internal ? 1 : 0
```
feeblefakie, line 65: Should it be created eve if `resource_count` is zero ?

modules/azure/scalardl/output.tf
feeblefakie: It's not working ?

modules/azure/scalardl/envoy.tf
tei-k: I think `envoy_public_ip` just for public lb. So fixed to use `enable_nlb` instead of `resource_count` like as aws.

modules/azure/scalardl/output.tf
tei-k: Yes, It didn't work. 🙄

feeblefakie: Then, I think it should be noted in the code.
If it is merged, it looks like it was working before.

feeblefakie: I got it !

PR #155, opened 2020-06-30
Refactoring docker role

tei-k:
# Description
https://scalar-labs.atlassian.net/browse/DLT-6232

# Done
- Refactoring to use `package` module to install `docker`.
- Delete yum-utils
- `docker-compose` and `cadvisor` can specify install version.
- Upgrade default version
  - `docker-compose` : 1.23.2 -> 1.26.0
  - `cadvisor` : 0.33.0 -> 0.35.0

# Didn't done
- Just support `CentOS` only (Haven't checked `Debian` or any other OS yet.)

# Confirm
```
[centos@scalardl-blue-1 ~]$ docker --version
Docker version 19.03.12, build 48a66213fe
[centos@scalardl-blue-1 ~]$ docker-compose --version
docker-compose version 1.26.0, build d4451659
[centos@scalardl-blue-1 ~]$ cadvisor --version
cAdvisor version v0.35.0 (49033161)

[centos@scalardl-blue-1 ~]$ docker ps
CONTAINER ID        IMAGE                           COMMAND                  CREATED             STATUS              PORTS                                  NAMES
5b142967228c        scalar-ledger-provision:2.0.4   "sh -c 'dockerize  -…"   48 minutes ago      Up 48 minutes       0.0.0.0:50051-50053->50051-50053/tcp   provision_scalar_1

[centos@scalardl-blue-1 ~]$ systemctl status cadvisor
● cadvisor.service - cAdvisor
   Loaded: loaded (/etc/systemd/system/cadvisor.service; enabled; vendor preset: disabled)
   Active: active (running) since Wed 2020-07-01 02:09:45 UTC; 49min ago
 Main PID: 7614 (cadvisor)
    Tasks: 12
   Memory: 26.7M
   CGroup: /system.slice/cadvisor.service
           └─7614 /usr/local/bin/cadvisor -port 18080

Jul 01 02:09:45 scalardl-blue-1 systemd[1]: Started cAdvisor.

[centos@scalardl-blue-1 bin]$ ll
total 45764
lrwxrwxrwx. 1 root root       31 Jul  1 07:01 cadvisor -> /usr/local/bin/cadvisor-v0.35.0
-rwxr-xr-x. 1 root root 34578680 Jul  1 07:01 cadvisor-v0.35.0
lrwxrwxrwx. 1 root root       36 Jul  1 06:45 docker-compose -> /usr/local/bin/docker-compose-1.26.1
-rwxr-xr-x. 1 root root 12255808 Jul  1 06:45 docker-compose-1.26.1
-rwxr-xr-x. 1 root root      200 Jul  1 06:43 j2
-rwxr-xr-x. 1 root root      365 Jul  1 07:00 pip
-rwxr-xr-x. 1 root root      367 Jul  1 07:00 pip3
-rwxr-xr-x. 1 root root      225 Jul  1 06:43 pip3.6
-rwxr-xr-x. 1 root root      371 Jul  1 07:00 pip3.8
```


provision/ansible/playbooks/roles/docker/tasks/cadvisor.yml
```diff
@@ -0,0 +1,35 @@
...
+    recurse: yes
+    state: directory
+  loop:
+    - /opt/cadvisor
+
+- name: Download cadvisor
+  get_url:
+    url: "{{ cadvisor_download }}"
+    checksum: sha256:{{ cadvisor_checksum }}
+    dest: "/opt/cadvisor/cadvisor-{{ cadvisor_version }}"
```
ymorimo, line 13: Is there any reason why cAdvisor is downloaded to `/opt` while docker-compose is downloaded to `/usr/local/bin` directly? Also, the version of docker-compose currently exists is checked but doesn't it do the same check for cAdvisor?
It would be nicer if the both tools are installed in the same way since they are both single binary. How about installing both tools like the following?
```
/usr/local/bin/docker-compose -> /usr/local/bin/docker-compose-{{ docker_compose_version }}
/usr/local/bin/cadvisor -> /usr/local/bin/cadvisor-{{ cadvisor_version }}
```

tei-k: Not for any particular reason, I'm just referenced community role.
https://github.com/nickjj/ansible-docker
https://github.com/ome/ansible-role-cadvisor

I agree with your suggestion. 👍 
Fixed in 02c6e87 (also upgrade the default version to 1.26.1 for docker-compose )

ymorimo: It got simpler 😀

provision/ansible/playbooks/roles/docker/handlers/main.yml
```diff
@@ -0,0 +1,12 @@
+---
+- name: Restart docker
+  systemd:
+    name: docker
+    state: restarted
+  become: yes
```
penpyt, line 6: [Q] I saw that you use a lot `become: yes` in almost every step, why not set `become: yes` in the playbook as follow?

```
- name: Docker Server
  hosts: all
  roles:
    - "docker"
  gather_facts: yes
  become: yes
```

tei-k: Good point! 👍 

Because the `ansible_user_id` variable becomes root when set the `become: yes` like as above.
Do you have a good idea?

```
https://github.com/scalar-labs/scalar-terraform/blob/6d0817c6e395871ae5205eeeda0fd530ef1bb15b/provision/ansible/playbooks/roles/docker/tasks/main.yml#L21-L26
```

penpyt: @tei-k if we do it in three-steps:
- get ansible_user_id with a become: no => modules setup
- set_facts to var
- adding an existing user to the docker group
what do you think?

feeblefakie: @tei-k @penpyt It can be merged but what do you guys want to do for this ?

tei-k: @penpyt Thanks!

It looks good,
But I think we should consider about docker group users. (it's not even within the scope of this PR.) 🙏

PR #156, opened 2020-07-02
Encrypt ebs volume for aws

tei-k:
# Description

Encrypt ebs volume

# Done
- cassandra data volume
- cassandra commit log volume
- monitor log volume


modules/aws/cassandra/cassandra.tf
```diff
@@ -41,12 +41,16 @@ module "cassandra_cluster" {
...
 
+data "aws_ebs_default_kms_key" "current" {}
+
 resource "aws_ebs_volume" "cassandra_data_volume" {
   count = local.cassandra.enable_data_volume && ! local.cassandra.data_use_local_volume ? local.cassandra.resource_count : 0
 
   availability_zone = local.locations[count.index % length(local.locations)]
   size              = local.cassandra.data_remote_volume_size
   type              = local.cassandra.data_remote_volume_type
+  encrypted         = true
```
feeblefakie, line 52: Sorry to say this now. 
The default should be `encrypted =true` but can you make it configurable ?

tei-k: >The default should be encrypted =true but can you make it configurable ?

OK! Fixed in 1abcd97. (Also deleted `kms_key_id` to use default kms_key - Just my check miss)

modules/aws/cassandra/locals.tf
```diff
@@ -22,6 +22,7 @@ locals {
     resource_type                = "t3.large"
     resource_count               = 3
     resource_root_volume_size    = 64
+    encrypt_ebs_volume           = true
```
feeblefakie, line 25:
```suggestion
    encrypt_volume           = true
```
so that we can use the same name for Azure.

tei-k: Fixed in cf2c465

PR #157, opened 2020-07-07
Refactoring common ansible role

tei-k:
# Description

https://scalar-labs.atlassian.net/browse/DLT-6566

# Done
- Fix to use `package` module to install python3 instead of `python_install.sh`.
- Delete `python_install.sh`

# Remark
I will separate `node-exporter` role from `common` role and refactoring in another PR.


provision/ansible/playbooks/roles/common/tasks/setup-redhat.yml
```diff
@@ -0,0 +1,14 @@
...
+    name: epel-release
+    state: present
+
+- name: Install python (redhat)
+  package:
+    name: "{{ item }}"
+    state: present
+  loop:
+    - python3
+    - python3-devel
```
ymorimo, line 13: Is `python3-devel` needed?

provision/ansible/playbooks/bastion-server.yml
```diff
@@ -8,7 +8,7 @@
       when: enable_tdagent is undefined or enable_tdagent == '1'
   vars:
     tdagent_conf_template: templates/bastion-server/td-agent.conf.j2
-  gather_facts: no
+  gather_facts: yes
```
feeblefakie, line 11: Any reason of this change ? I'm just wondering if it is required or just nice to have.

provision/ansible/playbooks/roles/common/tasks/setup-redhat.yml
tei-k: Good point! 👍 

Some pip packages require it and many python roles seem to often include it.
But it doesn't seem to be necessary now, so deleted in 3f06290.

https://github.com/mdklatt/ansible-python3-role/blob/df1cd558e24ea7f68edae37e9aae83cebf1a432a/tasks/centos.yml#L10

provision/ansible/playbooks/bastion-server.yml
tei-k: Because I need get `ansible_os_family` env using `gether_facts`.

```
when: ansible_os_family == 'RedHat'
```

tei-k:
>could we add the following package libselinux-python3 to let ansible copy file ?

@penpyt 

`libselinux-python` is installed, Does you mean need `libselinux-python3` ?
I'm just refactored `python_install.sh`, If you need this package, let we do it in another PR.

from `python_install.sh`
```
yum install -y python36 python36-setuptools libselinux-python
```


```
[centos@bastion-1 ~]$ rpm -qa |egrep "python3|libselinux"
python3-libs-3.6.8-13.el7.x86_64
python3-pip-9.0.3-7.el7_7.noarch
libselinux-2.5-14.1.el7.x86_64
python3-setuptools-39.2.0-10.el7.noarch
python3-3.6.8-13.el7.x86_64
libselinux-python-2.5-14.1.el7.x86_64
libselinux-utils-2.5-14.1.el7.x86_64
```

feeblefakie:
@penpyt @ymorimo I'm merging it but please feel free to ask.

penpyt:
> > could we add the following package libselinux-python3 to let ansible copy file ?
> 
> @penpyt
> 
> `libselinux-python` is installed, Do you mean need `libselinux-python3` ?
> I'm just refactored `python_install.sh`, If you need this package, let we do it in another PR.

I mean `libselinux-python` is for pyrthon2, where now we should use python3 with ansible as python2 is end-of-life, so we should install `libselinux-python3` to be able to use ansible_python_interpreter: "/usr/bin/python3".

tei-k:
>where now we should use python3 with ansible as python2 is end-of-life

I know this issue. I think it should be fix in other PR.

feeblefakie:
@penpyt @tei-k Let's work on the python upgrade in another PR since it is just a refactoring of the existing code without changing what it does.
BTW, is cqlsh is now working with python3 ?

penpyt:
> @penpyt @tei-k Let's work on the python upgrade in another PR since it is just a refactoring of the existing code without changing what it does. (We should separate non-directly related PRs if we can)
> BTW, is cqlsh is now working with python3 ?

@feeblefakie  cqlsh is only compatible with python2 https://docs.datastax.com/en/install/6.7/install/installCqlsh.html

PR #164, opened 2020-07-10
Refactoring td agent role

tei-k:
# Description
https://scalar-labs.atlassian.net/browse/DLT-6567

# Done
- Fix to use package module to install td-agent
- Create `conf.d` directory to include `*.conf`


provision/ansible/playbooks/bastion-server.yml
```diff
@@ -8,7 +8,11 @@
     - role: td-agent
       when: enable_tdagent is undefined or enable_tdagent == '1'
   vars:
-    tdagent_conf_template: templates/bastion-server/td-agent.conf.j2
+    tdagent_confd_templates:
+      - {
+          src: "templates/td-agent-confd/td-agent-bastion.conf.j2",
+          dest: "bastion.conf",
+        }
```
ymorimo, line 15: This is the same as the following, but any reason to format like this?

```yaml
    tdagent_confd_templates:
      - src: "templates/td-agent-confd/td-agent-bastion.conf.j2"
        dest: "bastion.conf"
```

tei-k: I just want to write in one line, But automatically formatted by my vscode plugin. 😓 
```
tdagent_confd_templates:
  - { src: "templates/td-agent-confd/td-agent-bastion.conf.j2", dest: "bastion.conf" }
```

Fixed in oneline (9f55901) How about this?

ymorimo: Looks good, but isn't it likely to be formatted again?

feeblefakie: I think what @ymorimo suggested [here](https://github.com/scalar-labs/scalar-terraform/pull/164#discussion_r454059067) is more common format of yaml.
@tei-k Can we do just make it like that ?

tei-k: Just disabled my plugin.
refs:
https://github.com/lrk/ansible-role-tdagent

feeblefakie: Hmm, not sure if it's good to disable that...

PR #168, opened 2020-07-22
Add docs about setting permissions

ymorimo:
https://scalar-labs.atlassian.net/browse/DLT-6137
https://scalar-labs.atlassian.net/browse/DLT-6136


docs/CloudPermissions.md
```diff
@@ -0,0 +1,130 @@
...
+            ],
+            "Resource": "*"
+        }
+    ]
+}
+```
+
+### Using S3 as a Terraform Backend
+
+If you use S3 as a Terraform backend, the action `s3:PutObject` is needed additionally.
```
feeblefakie, line 84:
```suggestion
If you use S3 as a Terraform backend, the action `s3:PutObject` needs to be added to the above list.
```

docs/CloudPermissions.md
```diff
@@ -0,0 +1,130 @@
...
+                "notDataActions": []
+            }
+        ]
+    }
+}
+```
+
+### Using Azure Storage as a Terraform Backend
+
+If you use Azure Storage as a Terraform backend, the built-in role `Storage Account Contributor` will work.
```
feeblefakie, line 130:
```suggestion
If you use Azure Storage as a Terraform backend, the built-in role `Storage Account Contributor` needs to be assigned additionally.
```

docs/CloudPermissions.md
```diff
@@ -0,0 +1,130 @@
...
+
+This guide explains the privileges for users to run scalar-terraform to deploy resources to the cloud.
+
+In general, users who have policies or roles with higher enough privileges, such as the `*FullAccess` actions on AWS or the `Contributor` role on Azure, can deploy resources with scalar-terraform without problems. But for more secure use, you can create a dedicated user and give it limited privileges.
+
+## AWS
+
+### General Policy for Deployment with scalar-terraform
+
+The following JSON is a general policy that allows users to manage resources for Scalar DLT with scalar-terraform.
```
feeblefakie, line 11: Just a question. What do you mean by `general` here ? I'm just wondering if we need it.

docs/CloudPermissions.md
```diff
@@ -0,0 +1,130 @@
+# Setting Cloud Permissions for scalar-terraform
+
+This guide explains the privileges for users to run scalar-terraform to deploy resources to the cloud.
+
+In general, users who have policies or roles with higher enough privileges, such as the `*FullAccess` actions on AWS or the `Contributor` role on Azure, can deploy resources with scalar-terraform without problems. But for more secure use, you can create a dedicated user and give it limited privileges.
```
feeblefakie, line 5:
```suggestion
In general from security perspective, it is better to assign restricted privileges to users not to make them able to do operations that they are not supposed to do.
```

docs/CloudPermissions.md
```diff
@@ -0,0 +1,130 @@
+# Setting Cloud Permissions for scalar-terraform
```
feeblefakie, line 1:
```suggestion
# Restricting Cloud Privileges for scalar-terraform
```

feeblefakie: Permission is also fine but I think we should use permission or privilege consistently.

ymorimo: Should the filename also be changed from `CloudPermissions.md` to `CloudPrivileges.md`?

docs/CloudPermissions.md
ymorimo: The policy lists a lot of permissions and doesn't limit the target resource (`"Resource": "*"`), so it gives users a very high privilege. There might be cases where you need to define more controlled and granular policies.

docs/CloudPermissions.md
feeblefakie: Yes, that would be better !

docs/CloudPermissions.md
feeblefakie: Ok, so maybe you mean, it is sufficient but not necessary ?
Then, for the title we can say something like `A Policy for ...` and leave a note about it as suggested below.

docs/CloudPermissions.md
```diff
@@ -0,0 +1,130 @@
...
+
+This guide explains how to restrict privileges for running scalar-terraform.
+
+In general from security perspective, it is better to assign restricted privileges to users not to make them able to do operations that they are not supposed to do.
+
+## AWS
+
+### General Policy for Deployment with scalar-terraform
+
+The following JSON is a general policy that allows users to manage resources for Scalar DLT with scalar-terraform.
```
feeblefakie, line 11: Something like this ?
```suggestion
The following JSON is a policy that allows users to manage resources for Scalar DLT with scalar-terraform. Please note that it is sufficient but not necessary since, for example, it doesn't limit the target resource with `"Resource": "*"` so it can/should be further restricted.
```

docs/CloudPermissions.md
ymorimo: Yes, so I will update the title and add the note as suggested. Thanks!

docs/CloudPrivileges.md
```diff
@@ -0,0 +1,130 @@
...
+
+### Using S3 as a Terraform Backend
+
+If you use S3 as a Terraform backend, the action `s3:PutObject` needs to be added to the above list.
+
+## Azure
+
+### A Role for Deployment with scalar-terraform
+
+The following JSON is a custom role that allow users to manage resources for Scalar DLT with scalar-terraform.
```
feeblefakie, line 90:
```suggestion
The following JSON is a custom role that allow users to manage resources for Scalar DLT with scalar-terraform. This it also sufficient but not necessary as described in AWS section.
```

PR #175, opened 2020-07-29
Fix unused volume bug

tei-k:
# Description
https://scalar-labs.atlassian.net/browse/DLT-6571

# Done
- Set default size for commit log volume.
- `bai` and `chiku` will create ebs volume and to use it.
- `sho` will use instance storage. (No change)


examples/aws/cassandra/example.tfvars
```diff
@@ -12,7 +12,7 @@ cassandra = {
   # data_remote_volume_size      = "64"
   # enable_commitlog_volume      = "false"
   # commitlog_use_local_volume   = "false"
-  # commitlog_remote_volume_size = ""
+  # commitlog_remote_volume_size = "32"
```
tei-k, line 15: Just default size.

modules/aws/cassandra/locals.tf
```diff
@@ -43,31 +43,28 @@ locals {
...
-        resource_type              = "r5d.large"
-        enable_data_volume         = true
-        data_remote_volume_size    = 1024
-        enable_commitlog_volume    = true
-        commitlog_use_local_volume = true
+        resource_type                = "r5d.large"
+        enable_data_volume           = true
+        data_remote_volume_size      = 1024
+        enable_commitlog_volume      = true
+        commitlog_remote_volume_size = 128
```
tei-k, line 50: `128` is referenced from bellow.
https://github.com/scalar-labs/scalardl-orchestration/blob/c03e984e12af56ce98f488053b7f245dbefebafe/provider/aws/cassandra/vars.tf#L95

modules/aws/cassandra/locals.tf
```diff
@@ -43,31 +43,28 @@ locals {
...
       }
     )
 
     sho = merge(local.cassandra_default,
       {
         resource_type              = "i3.2xlarge"
-        enable_data_volume         = true
         data_use_local_volume      = true
```
tei-k, line 67: I'm not sure why the data_volume uses the local volume (instance store storage - high performance but not persistent ).
This is how it's implemented since the old tools.

feeblefakie:
@tei-k Based on the discussion, it is not needed. Sorry for the confusion. Closing.

PR #181, opened 2020-08-12
Fix location to region for azure

tei-k:
# Description
Unify the names of variables with `region` and `locations` as well as aws.

https://docs.microsoft.com/en-us/azure/architecture/aws-professional/regions-zones

# Done
Fix var name from `location` to `region`.


ymorimo:
@tei-k `location` in the Inputs sections in READMEs of network and scalardl/cluster should also be updated?
* https://github.com/scalar-labs/scalar-terraform/blob/369cf20bdcd927a3361cffdb50ac9d86180e399f/modules/azure/network/README.md#inputs
* https://github.com/scalar-labs/scalar-terraform/blob/369cf20bdcd927a3361cffdb50ac9d86180e399f/modules/azure/scalardl/cluster/README.md#inputs

tei-k:
>location in the Inputs sections in READMEs of network and scalardl/cluster should also be updated?

@ymorimo Thanks! 👍  Updated in a774049.

PR #185, opened 2020-08-14
Update universal/scalardl module to use schema loader image for Cassandra

ymorimo:
https://scalar-labs.atlassian.net/browse/DLT-6879

This PR changes `universal/scalardl` module to use [scalardl-schama-loader-cassandra](https://github.com/scalar-labs/scalardl-schema-loader-cassandra) for the initial schema creation. The image is needed to be transferred to one of the scalardl hosts in the same way as the `scalar-ledger` image.

The `cqlsh` and `create_schema.cql` script in scalar-ledger image is no longer used. (They are going to be removed in the next version of scalar-ledger.)

### DIFF
--- a/modules/universal/scalardl/main.tf
+++ b/modules/universal/scalardl/main.tf
@@ -2,6 +2,11 @@ locals {
   scalar_image          = "${var.scalardl_image_name}:${var.scalardl_image_tag}"
   image_filename        = "${basename(var.scalardl_image_name)}-${var.scalardl_image_tag}.tar.gz"
   scalar_cassandra_host = "cassandra-lb.${var.internal_domain}"
+
+  schema_loader_cassandra_image_name     = "scalarlabs/scalardl-schema-loader-cassandra"
+  schema_loader_cassandra_image_tag      = "1.0.0"
+  schema_loader_cassandra_image          = "${local.schema_loader_cassandra_image_name}:${local.schema_loader_cassandra_image_tag}"
+  schema_loader_cassandra_image_filename = "${basename(local.schema_loader_cassandra_image_name)}-${local.schema_loader_cassandra_image_tag}.tar.gz"
 }
 
 module "ansible" {
@@ -13,6 +18,7 @@ resource "null_resource" "scalardl_image" {
 
   triggers = {
     scalar_tag = var.scalardl_image_tag
+    triggers   = join(",", var.triggers)
   }
 
   provisioner "local-exec" {
@@ -27,7 +33,6 @@ resource "null_resource" "scalardl_image_push" {
 
   triggers = {
     docker_image = null_resource.scalardl_image[0].id
-    triggers     = join(",", var.triggers)
   }
 
   connection {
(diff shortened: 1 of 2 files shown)


modules/universal/scalardl/main.tf
```diff
@@ -13,6 +16,7 @@ resource "null_resource" "scalardl_image" {
 
   triggers = {
     scalar_tag = var.scalardl_image_tag
+    triggers   = join(",", var.triggers)
```
tei-k, line 19: Why add `triggers` here ? Do we need to rebuild the image locally due to changes in bastion or Cassandra?

modules/universal/scalardl/main.tf
```diff
@@ -27,7 +31,6 @@ resource "null_resource" "scalardl_image_push" {
 
   triggers = {
     docker_image = null_resource.scalardl_image[0].id
-    triggers     = join(",", var.triggers)
```
tei-k, line 30: Is it related to this PR?
Maybe this is necessary, because if the bastion is replaced, we need to upload again to bastion.

modules/universal/scalardl/main.tf
```diff
@@ -148,7 +191,7 @@ resource "null_resource" "scalardl_schema" {
 
   provisioner "remote-exec" {
     inline = [
-      "docker run -e CASSANDRA_REPLICATION_FACTOR=${var.replication_factor} --rm ${local.scalar_image} dockerize -template create_schema.cql.tmpl:create_schema.cql -wait tcp://${local.scalar_cassandra_host}:9042 -timeout 30s cqlsh --cqlversion=3.4.4 ${local.scalar_cassandra_host} -u '${var.cassandra_username}' -p '${var.cassandra_password}' -f create_schema.cql"
+      "docker run -e CASSANDRA_HOST='${local.scalar_cassandra_host}' -e CASSANDRA_USERNAME='${var.cassandra_username}' -e CASSANDRA_PASSWORD='${var.cassandra_password}' -e CASSANDRA_REPLICATION_FACTOR=${var.replication_factor} --rm ${local.schema_loader_cassandra_image}"
```
tei-k, line 194: Why use single quotes? for escaping? 
I'm a little concerned when the password contains single quotes.

ymorimo: It seems that single quotations are used because they are surrounded by double quotations. The problem is the same as before. I will check if it fails if the username or password contains any single quotation, and report it as a bug lator.

modules/universal/scalardl/main.tf
ymorimo: I intended to modify the "triggers" tree  but I'm bit confused with it. I'd like to discuss this later.

tei-k: Thanks! I understand that this is an improvement on the triggers. 👍

modules/universal/scalardl/main.tf
tei-k: Thanks! 
It's an existing problem, and I think it's fine to consider it separately.

PR #195, opened 2020-09-02
Modification related to EBS data volume replacement

scalar-boney:
Cassandra node replacement doc updated for EBS data volume crash recovery.

- Cassandra node replacement with taint volume (Need to remove newly created node IP from seeds before restarting the new node).

- Replace the seeds of all Cassandra nodes after Cassandra node replacement(Replace newly created node IP).

### DIFF
--- a/docs/CassandraOperation.md
+++ b/docs/CassandraOperation.md
@@ -128,7 +128,7 @@ You need to modify the file `/etc/cassandra/conf/cassandra-env.sh` to add the li
 JVM_OPTS="$JVM_OPTS -Dcassandra.replace_address_first_boot=<dead_node_ip>"
 ```
 
-You also need to modify the file `/etc/cassandra/conf/cassandra.yaml` If you relaced a seed node and a volume.
+You also need to modify the file `/etc/cassandra/conf/cassandra.yaml` If you replaced a seed node and a volume.
 * Remove the newly created Cassandra node IP from the seeds of the node.
 
 Finally you can start the Cassandra service.
@@ -137,9 +137,7 @@ Finally you can start the Cassandra service.
 sudo systemctl start cassandra
 ```
 
-* After a successful restart of the newly created Cassandra node, you should replace the seeds of all Cassandra nodes with newly created Cassandra node IP and restart the Cassandra nodes.
-```console
-```
+If you replaced a seed node, you should replace the seeds of all Cassandra nodes with newly created Cassandra node IP.
 
 # Related Documents
 
--- a/docs/CassandraOperation.md
+++ b/docs/CassandraOperation.md
@@ -128,11 +128,16 @@ You need to modify the file `/etc/cassandra/conf/cassandra-env.sh` to add the li
 JVM_OPTS="$JVM_OPTS -Dcassandra.replace_address_first_boot=<dead_node_ip>"
 ```
 
+* You need to modify the file `/etc/cassandra/conf/cassandra.yaml` If you chose *Taint Volume*
+  * Remove the newly created cassandra node IP from the newly created cassandra node seeds.
+
 * Finally you can start the Cassandra service.
 
 ```console
 sudo systemctl start cassandra
 ```
+
+* After a successful restart of the newly created Cassandra node, you should replace the seeds of all Cassandra nodes with newly created Cassandra node IP.
 
 # Related Documents
 
--- a/docs/CassandraOperation.md
+++ b/docs/CassandraOperation.md
@@ -129,7 +129,7 @@ JVM_OPTS="$JVM_OPTS -Dcassandra.replace_address_first_boot=<dead_node_ip>"
 ```
 
 You also need to modify the file `/etc/cassandra/conf/cassandra.yaml` If you relaced a seed node and a volume.
-  * Remove the newly created cassandra node IP from the newly created cassandra node seeds.
+* Remove the newly created Cassandra node IP from the seeds of the node.
 
 * Finally you can start the Cassandra service.
 
--- a/docs/CassandraOperation.md
+++ b/docs/CassandraOperation.md
@@ -128,7 +128,7 @@ You need to modify the file `/etc/cassandra/conf/cassandra-env.sh` to add the li
 JVM_OPTS="$JVM_OPTS -Dcassandra.replace_address_first_boot=<dead_node_ip>"
 ```
 
-* You need to modify the file `/etc/cassandra/conf/cassandra.yaml` If you choose *Taint Volume*
+You also need to modify the file `/etc/cassandra/conf/cassandra.yaml` If you relaced a seed node and a volume.
   * Remove the newly created cassandra node IP from the newly created cassandra node seeds.
 
 * Finally you can start the Cassandra service.
--- a/docs/CassandraOperation.md
+++ b/docs/CassandraOperation.md
@@ -137,7 +137,7 @@ Finally you can start the Cassandra service.
 sudo systemctl start cassandra
 ```
 
-If you replaced a seed node, you should replace the seeds of all Cassandra nodes with newly created Cassandra node IP.
+If you replaced a seed node (old seed node), you should replace the old seed node's IP of all Cassandra nodes with newly created Cassandra node IP.
 
 # Related Documents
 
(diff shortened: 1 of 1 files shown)


docs/CassandraOperation.md
```diff
@@ -128,12 +128,20 @@ You need to modify the file `/etc/cassandra/conf/cassandra-env.sh` to add the li
 JVM_OPTS="$JVM_OPTS -Dcassandra.replace_address_first_boot=<dead_node_ip>"
 ```
 
+* You need to modify the file `/etc/cassandra/conf/cassandra.yaml` If you choose *Taint Volume*
+  * Remove the newly created cassandra node IP from the newly created cassandra node seeds.
```
feeblefakie, line 132:
```suggestion
* Remove the newly created Cassandra node IP from the seeds of the node.
```

docs/CassandraOperation.md
```diff
@@ -128,12 +128,20 @@ You need to modify the file `/etc/cassandra/conf/cassandra-env.sh` to add the li
...
+
 * Finally you can start the Cassandra service.
 
 ```console
 sudo systemctl start cassandra
 ```
 
+* After a successful restart of the newly created Cassandra node, you should replace the seeds of all Cassandra nodes with newly created Cassandra node IP and restart the Cassandra nodes.
+```console
+sudo systemctl restart cassandra
```
feeblefakie, line 142:
```suggestion
```

docs/CassandraOperation.md
```diff
@@ -128,12 +128,20 @@ You need to modify the file `/etc/cassandra/conf/cassandra-env.sh` to add the li
 JVM_OPTS="$JVM_OPTS -Dcassandra.replace_address_first_boot=<dead_node_ip>"
 ```
 
+* You need to modify the file `/etc/cassandra/conf/cassandra.yaml` If you choose *Taint Volume*
+  * Remove the newly created cassandra node IP from the newly created cassandra node seeds.
+
 * Finally you can start the Cassandra service.
```
feeblefakie, line 134:
```suggestion
Finally you can start the Cassandra service.
```

docs/CassandraOperation.md
tei-k, line 132: [Q] just remove ip from the newly created node? or all each nodes if replace node is seed node?

https://docs.datastax.com/en/ddac/doc/datastax_enterprise/operations/opsReplaceNode.html

feeblefakie (changes requested):

docs/CassandraOperation.md
scalar-boney: @tei-k san,
> just remove ip from the newly created node? or all each nodes if replaced node was seed node?

Before restarting the newly created node, We need to remove the newly created node IP from the `seeds` of newly created node.
Note:- It is applicable only for seed nodes. We can skip this step when the newly created node is a `non-seed` node.

docs/CassandraOperation.md
```diff
@@ -128,12 +128,20 @@ You need to modify the file `/etc/cassandra/conf/cassandra-env.sh` to add the li
...
 
 ```console
 sudo systemctl start cassandra
 ```
 
+* After a successful restart of the newly created Cassandra node, you should replace the seeds of all Cassandra nodes with newly created Cassandra node IP and restart the Cassandra nodes.
```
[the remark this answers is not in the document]
scalar-boney: > What do you mean by this?

All other Cassandra node seeds contain IP address of the old node(crashed node) When we replace a seed node. So we need to replace the IP address in all other Cassandra node seeds.

> I think non replaced nodes don't have the new node IP in the seeds list, doesn't it?

yes, We do not need to update the seeds for non replaced nodes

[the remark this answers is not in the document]
scalar-boney: Thank you for the review comment.
Document updated with the above change.

docs/CassandraOperation.md
yito88: Just curious. Is there another seed in this case? Can a new node join the cluster without removing the IP from its `seeds` parameter if other seeds are alive?

docs/CassandraOperation.md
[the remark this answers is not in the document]
yito88: I think that we don't need to make the new node a seed node. The new node might not be able to be accessed because the new node will have a heavy load to sync the data of the cluster.

I have two scenarios to replace a seed node. I'm not sure which case you tested.
### A) A cluster has only one seed node. The seed node is dead and we need to replace the node.
- The cluster has no living seed node, so, a new node can't join the cluster.
- Before replacing, we need to promote an existing node to a seed node. (We can remove the dead seed node's IP from other nodes in this promotion). After this promotion, this case will become a case where replacing a non-seed node.

### B) A cluster has multiple seed nodes. One of the seed nodes is dead and we replace it.
- A new node can join the cluster as a non-seed node.

According to the [document](https://docs.datastax.com/en/cassandra-oss/2.1/cassandra/operations/opsReplaceNode.html#Replacingadeadseednode), DataStax recommends promoting a non-seed node before replacing the dead seed node.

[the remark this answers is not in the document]
yito88: Sorry, the document is too old.
The current [document](https://docs.datastax.com/en/cassandra-oss/3.x/cassandra/operations/opsReplaceNode.html) says that a new node joins the cluster as a new seed node.

docs/CassandraOperation.md
scalar-boney: > Is there another seed in this case? 

Yes, In this case, other two `seed` nodes are available.

> Can a new node join the cluster without removing the IP from its seeds parameter if other seeds are alive?

Do you mean, 
Is it possible to add a new node in the cluster while one seed node crashed in the cluster?

yito88: > Is it possible to add a new node in the cluster while one seed node crashed in the cluster?
No. Can a new node with which the seed node was replaced join to the cluster without removing the seed's IP?
You removed the seed's IP from the config file of the new node in your test. I'd like to know what happens in the bootstrapping when it is NOT removed.

scalar-boney: > Can a new node with which the seed node was replaced join to the cluster without removing the seed's IP? 

yes, But both nodes will show in the cluster. (Old node status is `DN`)

> I'm wondering whether the new node checks another seed or not.
You removed the seed's IP from the config file of the new node in your test. I'd like to know what happens in the bootstrapping when it is NOT removed.

Bootstrapping will happens while the seeds are not replaced.

yito88: The terraform module automatically adds the new node's IP to `seeds` in the config file. So, we need to remove the IP from the `seeds` explicitly and restart the new node with `replace_address_first_boot` option. I got it.

docs/CassandraOperation.md
```diff
@@ -128,12 +128,17 @@ You need to modify the file `/etc/cassandra/conf/cassandra-env.sh` to add the li
 JVM_OPTS="$JVM_OPTS -Dcassandra.replace_address_first_boot=<dead_node_ip>"
 ```
 
-* Finally you can start the Cassandra service.
+You also need to modify the file `/etc/cassandra/conf/cassandra.yaml` If you replaced a seed node and a volume.
```
feeblefakie, line 131:
```suggestion
You also need to modify the file `/etc/cassandra/conf/cassandra.yaml` if you replaced a seed node and a volume.
```

docs/CassandraOperation.md
```diff
@@ -128,12 +128,17 @@ You need to modify the file `/etc/cassandra/conf/cassandra-env.sh` to add the li
...
+* Remove the newly created Cassandra node IP from the seeds of the node.
+
+Finally you can start the Cassandra service.
 
 ```console
 sudo systemctl start cassandra
 ```
 
+If you replaced a seed node, you should replace the seeds of all Cassandra nodes with newly created Cassandra node IP.
```
feeblefakie, line 140:
```suggestion
If you replaced a seed node (old seed node), you should replace the old seed node's IP of all Cassandra nodes with newly created Cassandra node IP.
```

feeblefakie (approved):
If the suggestions are applied, LGTM.

tei-k (approved):
LGTM!

PR #196, opened 2020-09-04
Add log archive storage

tei-k:
# Description
https://scalar-labs.atlassian.net/browse/DLT-7006

# Done
- Add feature for log archive to cloud storage (S3 and Azure blob)
- The default is not to archive.
- The s3 bucket must be created in the same region as other resources.
- Just support AWS -> S3, Azure -> Azure blob

### DIFF
--- a/modules/azure/monitor/identity.tf
+++ b/modules/azure/monitor/identity.tf
@@ -1,0 +1,20 @@
+data "azurerm_subscription" "subscription" {}
+
+data "azurerm_role_definition" "contributor" {
+  name = "Storage Blob Data Contributor"
+}
+
+data "azurerm_storage_account" "monitor" {
+  count = local.enable_log_archive_storage && local.monitor.resource_count > 0 ? 1 : 0
+
+  resource_group_name = local.network_name
+  name                = local.log_archive_storage.storage_account
+}
+
+resource "azurerm_role_assignment" "monitor" {
+  count = local.enable_log_archive_storage && local.monitor.resource_count > 0 ? local.monitor.resource_count : 0
+
+  scope              = data.azurerm_storage_account.monitor[0].id
+  role_definition_id = "${data.azurerm_subscription.subscription.id}${data.azurerm_role_definition.contributor.id}"
+  principal_id       = module.monitor_cluster.vm_identities[count.index][0]["principal_id"]
+}
--- a/modules/azure/monitor/identity.tf
+++ b/modules/azure/monitor/identity.tf
@@ -1,9 +1,5 @@
 data "azurerm_subscription" "subscription" {}
 
-data "azurerm_role_definition" "contributor" {
-  name = "Storage Blob Data Contributor"
-}
-
 data "azurerm_storage_account" "monitor" {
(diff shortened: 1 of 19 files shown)


modules/azure/monitor/locals.tf
```diff
@@ -57,4 +58,9 @@ locals {
     local.monitor_base[var.base],
     var.monitor
   )
+
+  log_archive_storage = regex("^(?:https://(?P<storage_account>[^.]+)[^/]+(?:/(?P<container_name>[^.]+)))?", local.monitor.log_archive_storage_base_uri)
+
```
feeblefakie, line 63: Can we remove the empty line? (for consistency)

modules/universal/monitor/vars.tf
```diff
@@ -68,5 +68,10 @@ variable "targets" {
 
 variable "log_retention_period_days" {
   default     = 30
-  description = "Set the retention period of the aggregated log to the monitor server"
+  description = "Set the retention period of the aggregated log to the monitor host"
```
feeblefakie, line 71:
```suggestion
  description = "Set the retention period of the accumulated logs in the monitor host"
```

modules/universal/monitor/vars.tf
```diff
@@ -68,5 +68,10 @@ variable "targets" {
...
+}
+
+variable "log_archive_storage_info" {
+  default     = ""
+  description = "A info to set archive storage of the aggregated log to the monitor host"
```
feeblefakie, line 76:
```suggestion
  description = "An info to set archive storage of the accumulated logs in the monitor host"
```

It is not related to this PR but I think all the descriptions are a bit inconsistent in the sense that some start with a noun describing what it is and others start with a verb describing what it does.

feeblefakie (commented):
Overall looking good! Left minor comments.

modules/azure/monitor/locals.tf
tei-k: OK! Fixed in 2e5c892.

modules/universal/monitor/vars.tf
tei-k: Fixed in ff60449.

modules/universal/monitor/vars.tf
tei-k: Fixed in ff60449.

modules/azure/monitor/identity.tf
```diff
@@ -0,0 +1,32 @@
+data "azurerm_subscription" "subscription" {}
+
+data "azurerm_storage_account" "monitor" {
+  count = local.enable_log_archive_storage && local.monitor.resource_count > 0 ? 1 : 0
+
+  resource_group_name = local.network_name
+  name                = local.log_archive_storage.storage_account
+}
+
+resource "azurerm_role_definition" "monitor" {
```
ymorimo, line 10: This is good compared to Cassy's identity that uses the built-in `Storage Blob Data Contributor` role.

ymorimo (approved):
LGTM!

PR #201, opened 2020-09-10
Fix to create role definition only in enable log archive

tei-k:
# Description
https://scalar-labs.atlassian.net/browse/DLT-7189

Terratest failed because don't have permission to create the role definition.

# Done
Fix to create role definition only in enable log archive.


feeblefakie (approved):
LGTM!

feeblefakie:
@ymorimo I'm merging this but PTAL when you get a chance.

PR #202, opened 2020-09-11
Fix doc to use terraform state list

tei-k:
# Description

Since most resources start with `module.cassandra`, `grep` is not a smart way to list resources.

# Done
Fix to use `terraform state list`


feeblefakie (approved):
LGTM!

PR #220, opened 2020-10-01
Troubleshoot document updated

scalar-boney:
Troubleshoot document updated with a new condition.
- Replace accidentally deleted node or disk after initial configuration of shared environment

### DIFF
--- a/docs/TroubleshootingGuide.md
+++ b/docs/TroubleshootingGuide.md
@@ -17,22 +17,12 @@ If you accidentally delete a node that contains an additional data disk in Azure
 
 Please try the following
 * Delete the os-disk If the node is terminated.
-* Do `terraform state rm` as follows.
-  
+* Do `terraform refresh` or `terraform state rm` as follows.
 ```console
 terraform state rm "module.cassandra.module.cassandra_cluster.azurerm_virtual_machine.vm-linux[0]"
 terraform state rm "module.cassandra.azurerm_virtual_machine_data_disk_attachment.cassandra_data_volume_attachment[0]"
-
-terraform apply
 ```
+* Do `terraform apply`
 
 Follow [Cassandra Post Recovery Steps](CassandraOperation.md#post-recovery-steps)
 
-## Replace accidentally deleted node or disk after initial configuration of shared environment
-
-When an error is faced like `module does not exist` while recreating the accidentally deleted node or disk using `terraform state rm` or `terraform taint` command after configuring share environment in the local system, following workaround
-
-There is no need to execute `terraform state rm` or `terraform taint` command for node or disk replacement.
-Because terminated node or disk modules will be removed from `tfstate` file while executing `terraform refresh` command for replacing the keys in `tfstate` file as part of [ShareEnvironment](ShareEnvironment.md).
-
-So terminated node can be simply replaced with `terraform apply` command.
--- a/docs/TroubleshootingGuide.md
+++ b/docs/TroubleshootingGuide.md
@@ -30,9 +30,9 @@ Follow [Cassandra Post Recovery Steps](CassandraOperation.md#post-recovery-steps
 
 ## Replace accidentally deleted node or disk after initial configuration of shared environment
 
-When an error is faced like `module does not exist` while configuring shared environment in local system after a node or data disk was accidentally removed from scalar-terraform environment, following workaround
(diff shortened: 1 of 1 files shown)


docs/TroubleshootingGuide.md
```diff
@@ -27,3 +27,12 @@ terraform apply
 ```
 
 Follow [Cassandra Post Recovery Steps](CassandraOperation.md#post-recovery-steps)
+
+## Replace accidentally deleted node or disk after initial configuration of shared environment
+
+When an error is faced like `module does not exist` while configuring shared environment in local system after a node or data disk was accidentally removed from scalar-terraform environment, following workaround
```
tei-k, line 33: @scalar-boney 

Sorry! I'm not sure what kind of case this is.

>a node or data disk was accidentally removed from scalar-terraform environment

What kind of case was removed by accident?

tei-k: Maybe I understood what you're saying.

But I don't think we need to write about this case in troubleshooting, since we can recreate resources normally with `terraform apply`! 🤔 

If you're going to write it, I think you should make the conditions a little clearer.

docs/TroubleshootingGuide.md
```diff
@@ -27,3 +27,12 @@ terraform apply
...
+
+There is no need to execute `terraform state rm` or `terraform state taint` command for node or disk replacement.
```
tei-k, line 35: `terraform state taint` → `terraform taint` ? 

There doesn't seem to be `taint` as a sub command of `state`.

docs/TroubleshootingGuide.md
scalar-boney: Thank you for the review comments.
Doc modified with more clarification.

docs/TroubleshootingGuide.md
scalar-boney: Thank you for the detailed review.
Sorry, 
Doc updated.

docs/TroubleshootingGuide.md
```diff
@@ -27,3 +27,12 @@ terraform apply
 ```
 
 Follow [Cassandra Post Recovery Steps](CassandraOperation.md#post-recovery-steps)
+
+## Replace accidentally deleted node or disk after initial configuration of shared environment
+
+When an error is faced like `module does not exist` while recreating the accidentally deleted node or disk using `terraform state rm` or `terraform taint` command after configuring share environment in the local system, following workaround
+
+There is no need to execute `terraform state rm` or `terraform taint` command for node or disk replacement.
+Because terminated node or disk modules will be removed from `tfstate` file while executing `terraform refresh` command for replacing the keys in `tfstate` file as part of [ShareEnvironment](ShareEnvironment.md).
```
tei-k, line 36: In this case, I think `terraform refresh` could be used instead of `terraform state rm xxx`.
So either `terraform state rm xxx` or `terraform refresh` is the solution.

```
The terraform refresh command is used to reconcile the state Terraform knows about (via its state file) with the real-world infrastructure. This can be used to detect any drift from the last-known state, and to update the state file.
```
https://www.terraform.io/docs/commands/refresh.html

scalar-boney: Yes, That's right.

This is a rare case of terminated instance recovery.
In this section, I am trying to explain how to recover the terminated instance after the initial setup of the shared environment.

While we are configuring the shared environment [ShareEnvironment](ShareEnvironment.md), It will execute the `terraform refresh` command.

Sorry, Do you mean to modify the section `Recover a node with existing data disk from accidental deletion of a node in Azure` using `terraform refresh`?

tei-k: >Do you mean to modify the section Recover a node with existing data disk from accidental deletion of a node in Azure using terraform refresh?

Yes, In this case, I think both `refresh` and `state rm` are OK.

Maybe We can write both as below. 
So I don't think this part `## Replace accidentally deleted node or disk after initial configuration of shared environment` is necessary.

e.g.

>* Do `terraform state rm` as follows.
* Do `terraform refresh` or `terraform state rm` as follows.
・・・

scalar-boney: Thank you for your review comments.
Troubleshoot Guide has been revised accordingly.

tei-k (approved):
LGTM!

feeblefakie (approved):
LGTM

PR #224, opened 2020-10-09
Make monitor module use external volume always

ymorimo:
https://scalar-labs.atlassian.net/browse/DLT-6177

Removed the option `enable_log_volume` from the monitor module. An external volume should always be created to store logs.

### DIFF
--- a/modules/azure/monitor/main.tf
+++ b/modules/azure/monitor/main.tf
@@ -26,7 +26,7 @@ module "monitor_cluster" {
 }
 
 resource "azurerm_managed_disk" "monitor_log_volume" {
-  count = local.monitor.resource_count
+  count = local.monitor.enable_tdagent ? local.monitor.resource_count : 0
 
   name                 = "log-${count.index + 1}"
   location             = local.region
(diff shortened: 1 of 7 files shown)


modules/azure/monitor/main.tf
```diff
@@ -26,7 +26,7 @@ module "monitor_cluster" {
 }
 
 resource "azurerm_managed_disk" "monitor_log_volume" {
-  count = local.monitor.enable_log_volume ? local.monitor.resource_count : 0
+  count = local.monitor.resource_count
```
feeblefakie, line 29: Not checking if `enable_tdagent` is a bug of the Azure module?

feeblefakie (commented):
Thanks! Left 1 question.

modules/azure/monitor/main.tf
ymorimo: Ah, right. Updated to check the variable for the Azure module to match the AWS in this PR: 1c86ddb

tei-k (approved):
LGTM! 👍

feeblefakie (approved):
LGTM!

PR #230, opened 2020-10-20
Change default vm size for azure bastion

tei-k:
# Description
https://scalar-labs.atlassian.net/browse/DLT-7403

# Done
- Change default vm size from `Standard_D2s_v3` to `Standard_B1s`
- Change OS disk size to `32GB` (P4 tier - $6.08/mo)
```
Error: compute.VirtualMachinesClient#CreateOrUpdate: Failure sending request: StatusCode=0 -- Original Error: autorest/azure: Service returned an error. Status=<nil> Code="OperationNotAllowed" Message="The specified disk size 16 GB is smaller than the size of the corresponding disk in the VM image: 30 GB. This is not allowed. Please choose equal or greater size or do not specify an explicit size." Target="osDisk.diskSizeGB"
```


feeblefakie (approved):
LGTM! Thanks.

PR #188, opened 2020-08-21
New environment variable added in scalardl env file

scalar-boney:
New environment variables added in scalardl `env` file for creating schema manually using `scalardl-schema-loader-cassandra` image.

```
docker run -e CASSANDRA_HOST=${SCALAR_CASSANDRA_HOST} -e CASSANDRA_USERNAME=${SCALAR_CASSANDRA_USERNAME} -e CASSANDRA_PASSWORD=${SCALAR_CASSANDRA_PASSWORD} -e CASSANDRA_REPLICATION_FACTOR=${CASSANDRA_REPLICATION_FACTOR} --rm ${SCHEMA_LOADER_CASSANDRA_IMAGE}
```
Modification based on the review comments
https://github.com/scalar-labs/misc/pull/9#discussion_r474422508

PR #231, opened 2020-10-28
Replace scalardl-schema-loader-cassandra with scalardl-schema-loader

ymorimo:
https://scalar-labs.atlassian.net/browse/DLT-7405

This PR migrates`scalardl-schema-loader-cassandra` to [the unified `scalardl-schema-loader`](https://github.com/scalar-labs/scalardl-schema-loader/tree/v1.1.0) in scalar-terraform.

This is the first step to support DynamoDB. Note that the current `scalarlabs/scalardl-schema-loader:1.1.0` doesn't support DynamoDB yet, so it will be updated when the DynamoDB support is added.

### DIFF
--- a/modules/universal/scalardl/main.tf
+++ b/modules/universal/scalardl/main.tf
@@ -3,8 +3,7 @@ locals {
   image_filename        = "${basename(var.scalardl_image_name)}-${var.scalardl_image_tag}.tar.gz"
   scalar_cassandra_host = "cassandra-lb.${var.internal_domain}"
 
-  schema_loader_cassandra_image          = "${var.schema_loader_cassandra_image_name}:${var.schema_loader_cassandra_image_tag}"
-  schema_loader_cassandra_image_filename = "${basename(var.schema_loader_cassandra_image_name)}-${var.schema_loader_cassandra_image_tag}.tar.gz"
+  schema_loader_image_filename = format("%s%s", replace(basename(var.schema_loader_image), ":", "-"), ".tar.gz")
 }
 
 module "ansible" {
@@ -46,7 +45,7 @@ resource "null_resource" "scalardl_image_push" {
   }
 }
 
-resource "null_resource" "schema_loader_cassandra_image" {
+resource "null_resource" "schema_loader_image" {
   count = var.provision_count > 0 ? 1 : 0
 
   triggers = {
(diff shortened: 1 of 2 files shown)


feeblefakie (approved):
LGTM!

modules/universal/scalardl/main.tf
```diff
@@ -218,8 +217,6 @@ resource "null_resource" "scalardl_container" {
       "echo export SCALAR_CASSANDRA_HOST=${local.scalar_cassandra_host} >> env",
       "echo export SCALAR_CASSANDRA_USERNAME=${var.cassandra_username} >> env",
       "echo export SCALAR_CASSANDRA_PASSWORD=${var.cassandra_password} >> env",
-      "echo export SCHEMA_LOADER_CASSANDRA_IMAGE=${local.schema_loader_cassandra_image} >> env",
-      "echo export CASSANDRA_REPLICATION_FACTOR=${var.replication_factor} >> env",
```
tei-k, line 222: These schema loader info may still be need to run from ansible of manual-testing.
#188

ymorimo: Thank you for the info. I think the Ansible playbook in the scalar-labs/misc repo should be updated according to this change.

ymorimo: @tei-k How about updating [the README](https://github.com/scalar-labs/misc/blob/master/tools/manual-testing/README.md#script-7---create-new-scalar-schema) on the manual-testing side as follows? Cc: @scalar-boney

```
ansible-playbook create_scalar_schema.yml -e "host=scalardl-blue-1.internal.scalar-labs.com schema_loader_image=scalarlabs/scalardl-schema-loader:1.1.0 cassandra_replication_factor=3"
```

tei-k: @ymorimo 

It's looking good. 👍 
But just need to check the version of the schema loader image current using.
That's why to put its in here as environment variable.

Also need to fix the following prat. (I didn't check them all.)
https://github.com/scalar-labs/misc/blob/a320a55517dc53b393baaf08e8da338472565669/tools/manual-testing/ansible/create_scalar_schema.yml#L16

ymorimo: I opened a PR to modify the manual testing. scalar-labs/misc#12

tei-k (approved):

PR #233, opened 2020-11-05
Fix to use standard ssd for azure scalardl

tei-k:
# Description
https://scalar-labs.atlassian.net/browse/DLT-7571

# Done
- Fix to use `standard ssd` in envoy and scalardl VM in Azure.

# Ref
https://azure.microsoft.com/en-us/blog/preview-standard-ssd-disks-for-azure-virtual-machine-workloads/


feeblefakie (approved):
LGTM! Thank you!

PR #264, opened 2021-01-27
Fix to use standard load balancer for azure

tei-k:
# Description
https://scalar-labs.atlassian.net/browse/DLT-8030

`Microsoft recommends Standard load balancer`
https://docs.microsoft.com/en-us/azure/load-balancer/skus

# Done
Fix to use `Standard Load Balancer` in Azure.

### DIFF
--- a/modules/azure/scalardl/envoy.tf
+++ b/modules/azure/scalardl/envoy.tf
@@ -72,7 +72,7 @@ resource "azurerm_public_ip" "envoy_public_ip" {
   name                = "PublicIPForEnvoy"
   domain_name_label   = "envoy-${local.network_name}"
   location            = local.region
-  sku                 = length(local.locations) > 0 ? "Standard" : "Basic"
+  sku                 = "Standard"
   resource_group_name = local.network_name
   allocation_method   = "Static"
 }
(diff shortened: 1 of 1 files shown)


ymorimo (approved):
LGTM!

modules/azure/scalardl/envoy.tf
```diff
@@ -88,7 +88,7 @@ resource "azurerm_public_ip" "envoy_nat_ip" {
 }
 
 resource "azurerm_nat_gateway" "envoy_natgw" {
-  count = local.envoy.enable_nlb && local.envoy.nlb_internal && length(local.locations) > 0 ? 1 : 0
+  count = local.envoy.enable_nlb && local.envoy.nlb_internal ? 1 : 0
```
feeblefakie, line 91: Just a question.
Does this change make envoy nodes use NAT for going out if both are true?

feeblefakie (approved):
LGTM!

modules/azure/scalardl/envoy.tf
tei-k: Yes, If the `LB` is `Internal Standard SKU Load Balancer` and VM does not have  a public IP, use NAT for outbound connections.

PR #271, opened 2021-01-29
Disable autoscaling by defaulte for azure k8s

tei-k:
# Description

https://scalar-labs.atlassian.net/browse/DLT-8088

# Done

Fix `cluster_auto_scaling` to `false` by default.

PR #272, opened 2021-02-01
Fix autoscale in azure k8s

tei-k:
# Description

#271 

```
2021-01-31T15:47:45.6025359Z TestEndToEndK8s 2021-01-31T15:47:45Z command.go:158: Error: expanding `default_node_pool`: `max_count`(6) and `min_count`(3) must be set to `null` when `enable_auto_scaling` is set to `false`
```

# Done
Set to `null` when `enable_auto_scaling` is `false`.


feeblefakie (approved):
LGTM!

PR #273, opened 2021-02-03
Able to configure cluster accessibility for aks

tei-k:
# Description
https://scalar-labs.atlassian.net/browse/DLT-8250

# Done
- Add `public_cluster_enabled` var for configure cluster
- Fix to use `public_cluster_enabled` instead of `private_cluster_enabled`
`public_cluster_enabled` : `false` => `Private only`
`public_cluster_enabled` : `true` => `Public and Private`
https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/kubernetes_cluster#private_cluster_enabled
```
Should this Kubernetes Cluster have its API server only exposed on internal IP addresses? This provides a Private IP Address for the Kubernetes API on the Virtual Network where the Kubernetes Cluster is located. Defaults to false. Changing this forces a new resource to be created.
```


feeblefakie (approved):
LGTM!

ymorimo (approved):
LGTM!

PR #280, opened 2021-02-16
Remove unused mount point for tailon container

ymorimo:
A mount point for the config file of tailon container is not used since the file is not specified in the command line.


tei-k (approved):
LGTM! Thanks!

feeblefakie (approved):
LGTM!

PR #288, opened 2021-02-26
Fix scalardl instance size for aws chiku

tei-k:
# Description

https://scalar-labs.atlassian.net/browse/DLT-7847

# Done

before(`t3.large`) -> after(`t3.medium`)
📝 `Standard_B2s` (2 vcpus, 4GB mem) = `t3.medium` (2 vcpus, 4GB mem)

### DIFF
--- a/modules/aws/scalardl/locals.tf
+++ b/modules/aws/scalardl/locals.tf
@@ -51,9 +51,7 @@ locals {
 
     bai = merge(local.scalardl_default, {})
 
-    chiku = merge(local.scalardl_default,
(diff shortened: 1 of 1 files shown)


feeblefakie:
@tei-k I think `Standard_B2s` is a burst instance so we need to update the azure one also if we switch to t3.medium, which is not burst instance.

feeblefakie (approved):
LGTM!

ymorimo (commented):
LGTM!

PR #289, opened 2021-03-03
Output local files for k8s config and ssh config

tei-k:
# Description

Output files for `ssh config`, `inventory`, `kube config`


feeblefakie (approved):
LGTM!

feeblefakie:
@ymorimo Merging it but PTAL when you get a chance!

modules/azure/network/locals.tf
```diff
@@ -58,3 +58,36 @@ locals {
...
+[bastion]
+%{for f in module.bastion.bastion_host_ips~}
+${f}
+%{endfor}
+
+[bastion:vars]
+host=bastion
+
+[all:vars]
+internal_domain=${local.network.internal_domain}
```
ymorimo, line 87: @tei-k I got an error here.

```
Error: Unsupported attribute

  on .terraform/modules/network/modules/azure/network/locals.tf line 87, in locals:
  87: internal_domain=${local.network.internal_domain}
    |----------------
    | local.network is object with 9 attributes

This object does not have an attribute named "internal_domain".
```

Maybe `var.internal_domain` is correct?

tei-k: Thanks! fixed in #289

PR #290, opened 2021-03-03
Upgrade k8s version to 1.19

tei-k:


modules/azure/kubernetes/locals.tf
```diff
@@ -27,11 +27,11 @@ locals {
     resource_group_name       = var.network.name
     region                    = var.network.region
     dns_prefix                = "scalar-kubernetes"
-    kubernetes_version        = "1.16.13"
+    kubernetes_version        = "1.19.7"
     admin_username            = "azureuser"
     public_ssh_key_path       = var.network.public_key_path
     role_based_access_control = true
-    kube_dashboard            = true
+    kube_dashboard            = false
```
feeblefakie, line 34: Is this change from some limitation?
If so, we should note it here as a comment.

feeblefakie (commented):
Left one question.

modules/azure/kubernetes/locals.tf
tei-k: Because kube_dashboard add-on is deprecated for versions 1.19+, so deleted this option in e3c0d77.

feeblefakie (approved):
LGTM! Thank you!

feeblefakie:
@ymorimo Merging it but PTAL when you get a chance.

ymorimo (commented):
LGTM! Sorry for the late review.

PR #293, opened 2021-03-10
Add Scalar Auditor deployment support to scalardl module

ymorimo:
https://scalar-labs.atlassian.net/browse/DLT-8326

* Introduce the container variables file (`container_env_file`).
* Add port numbers variables for the `scalar-auditor` container to the docker-compose.yml in the `scalardl` module.


modules/universal/scalardl/main.tf
```diff
@@ -195,14 +216,14 @@ resource "null_resource" "scalardl_schema" {
...
     inline = [
-      format(
-        "%s %s",
-        "docker run --rm ${var.schema_loader_image} -h ${var.database_contact_points} -u ${var.database_username} -p ${var.database_password}",
-        var.database == "cassandra" ? "--cassandra -P ${var.database_contact_port} -n NetworkTopologyStrategy -R ${var.cassandra_replication_factor}" :
-        var.database == "dynamo" ? "--dynamo --region ${var.database_contact_points}" :
-        var.database == "cosmos" ? "--cosmos -h ${var.database_contact_points} -p ${var.database_password}" :
-        ""
-      )
+      "set -x",
```
ymorimo, line 219: `set -x` is useful with `remote-exec` to see what is going on in the provisioning process.

modules/universal/scalardl/main.tf
```diff
@@ -195,14 +216,15 @@ resource "null_resource" "scalardl_schema" {
...
+      # Source only SCALAR_DB_* variables from the env file for docker-compose
+      "source <(grep '^SCALAR_DB_' $HOME/provision/container.env)",
```
ymorimo, line 221: Source only `SCALAR_DB_*` vars here since `SCALAR_DL_*` vars can contain spaces which leads to a shell error. For example, the following line fails with an error `PRIVATE: command not found` when sourced.

```
SCALAR_DL_AUDITOR_PRIVATE_KEY_PEM=-----BEGIN PRIVATE KEY-----...
```

feeblefakie (dismissed):
LGTM! Thank you!

ymorimo:
@feeblefakie Thank you for the review, but it has been dismissed because I added a commit to fix the schema loading script.

feeblefakie (dismissed):
Still LGTM!

tei-k:
Why not fix Azure's module?

https://github.com/scalar-labs/scalar-terraform/blob/4673a62068c0acdb70f92860569dc41efe1673e9/modules/azure/scalardl/cluster/vars.tf#L104-L127

https://github.com/scalar-labs/scalar-terraform/blob/4673a62068c0acdb70f92860569dc41efe1673e9/modules/azure/scalardl/cluster/main.tf#L37-L41

ymorimo:
> Why not fix Azure's module?

I forgot to mention that, but  I'm adding the same changes for Azure in another PR once this PR is approved. I've updated the PR title.

ymorimo:
> > Why not fix Azure's module?
> 
> I forgot to mention that, but I'm adding the same changes for Azure in another PR once this PR is approved. I've updated the PR title.

Sorry, but this PR would prevent the modules for Azure from working. So I will append the changes for Azure to this PR.

ymorimo:
Added the same changes to the `azure/scalardl` module as the `aws/scalardl` module.

modules/aws/scalardl/locals.tf
```diff
@@ -26,16 +26,13 @@ locals {
     blue_image_tag               = "2.1.0"
     blue_image_name              = "ghcr.io/scalar-labs/scalar-ledger"
     blue_discoverable_by_envoy   = true
+    blue_container_env_file      = "scalardl_blue_container.env"
     green_resource_count         = 0
     green_image_tag              = "2.1.0"
     green_image_name             = "ghcr.io/scalar-labs/scalar-ledger"
     green_discoverable_by_envoy  = false
+    green_container_env_file     = "scalardl_green_container.env"
```
tei-k, line 34: Do we need to separate the environment files for blue and green?

modules/universal/scalardl/main.tf
```diff
@@ -173,16 +177,33 @@ resource "null_resource" "schema_loader_image_load" {
   }
 }
 
+resource "null_resource" "scalardl_container_env_file_push" {
+  count = var.provision_count > 0 ? 1 : 0
```
tei-k, line 181: Isn't this necessary for all nodes?
```suggestion
  count = var.provision_count
```

modules/universal/scalardl/main.tf
```diff
@@ -173,16 +177,33 @@ resource "null_resource" "schema_loader_image_load" {
...
+    bastion_host = var.bastion_host_ip
+    host         = var.host_list[count.index]
+    user         = var.user_name
+    agent        = true
+    private_key  = file(var.private_key_path)
+  }
+
+  provisioner "remote-exec" {
+    inline = ["mkdir -p $HOME/provision"]
+  }
```
tei-k, line 193: Is this necessary? it looks like created in `scalardl_load` too.

modules/universal/scalardl/main.tf
```diff
@@ -137,9 +137,13 @@ resource "null_resource" "scalardl_load" {
...
+
   provisioner "file" {
-    source      = "${path.module}/provision"
-    destination = "$HOME"
+    source      = "${path.module}/provision/"
+    destination = "$HOME/provision"
   }
```
tei-k, line 147: I'm not sure why this part needs to be fixed. is it a bug?

modules/universal/scalardl/main.tf
ymorimo: This is needed because the `provision` directory doesn't exist if `null_resource.scalardl_container_env_file_push` is run first.
By this line and [the change in `null_resource.scalardl_load`](https://github.com/scalar-labs/scalar-terraform/pull/293/files/55b04f4e686b053c915123e5b443918b8992c192#diff-f5f0ce77ada17f7d4a2c38d79fdb69f2b0bc27c8628307e7857bbe1402906ebeR145-R146), it is guaranteed that the `provision` directory is created or coped "as a directory" regardless of the order of `null_resource`s.

modules/universal/scalardl/main.tf
ymorimo: Answered in https://github.com/scalar-labs/scalar-terraform/pull/293#discussion_r604529256

modules/universal/scalardl/main.tf
tei-k: OK! But I think it would be better to use `triggers` to control the execution order. What do you think?

modules/aws/scalardl/locals.tf
ymorimo: Good point. The environment files might be shared with blue and green since the vars in this file are basically for the database.
@feeblefakie What do you think?

modules/universal/scalardl/main.tf
ymorimo: `scalardl_load` and `scalardl_container_env_file_push` should be able to be executed or re-executed by `taint`ing independently, so I don't think there are any dependencies on each other.

tei-k: Well. just for order.

Also need to add `null_resource.scalardl_container_env_file_push[count.index].id` to triggers.
https://github.com/scalar-labs/scalar-terraform/blob/4673a62068c0acdb70f92860569dc41efe1673e9/modules/universal/scalardl/main.tf#L213-L216

modules/universal/scalardl/main.tf
```diff
@@ -195,14 +216,15 @@ resource "null_resource" "scalardl_schema" {
...
-        "%s %s",
-        "docker run --rm ${var.schema_loader_image} -h ${var.database_contact_points} -u ${var.database_username} -p ${var.database_password}",
-        var.database == "cassandra" ? "--cassandra -P ${var.database_contact_port} -n NetworkTopologyStrategy -R ${var.cassandra_replication_factor}" :
-        var.database == "dynamo" ? "--dynamo --region ${var.database_contact_points}" :
-        var.database == "cosmos" ? "--cosmos -h ${var.database_contact_points} -p ${var.database_password}" :
-        ""
-      )
+      "set -x",
+      # Get the SCALAR_DB_* variables from the env file for docker-compose
+      "eval `grep '^SCALAR_DB_' $HOME/provision/container.env`",
```
tei-k, line 221: $(command) is recommended over backquote.
https://google.github.io/styleguide/shellguide.html#shellcheck

modules/universal/scalardl/main.tf
```diff
@@ -195,14 +216,15 @@ resource "null_resource" "scalardl_schema" {
...
+      "cmd=\"docker run --rm ${var.schema_loader_image} -u $SCALAR_DB_USERNAME -p $SCALAR_DB_PASSWORD\"",
+      "if [[ $SCALAR_DB_STORAGE == 'cassandra' ]]; then $cmd --cassandra -h $SCALAR_DB_CONTACT_POINTS -P $SCALAR_DB_CONTACT_PORT -n NetworkTopologyStrategy -R ${var.cassandra_replication_factor};",
+      "elif [[ $SCALAR_DB_STORAGE == 'dynamo' ]]; then $cmd --dynamo --region $SCALAR_DB_CONTACT_POINTS;",
+      "elif [[ $SCALAR_DB_STORAGE == 'cosmos' ]]; then $cmd --cosmos -h $SCALAR_DB_CONTACT_POINTS -p $SCALAR_DB_PASSWORD;",
```
tei-k, line 225: Is it okay to have two `-p $SCALAR_DB_PASSWORD` and a `-u $SCALAR_DB_USERNAME` in the case of `cosmos` ?

tei-k: maybe better to create a script file. (e.g., `xxx.sh schema_load_image env_file`)

tei-k:
Do we provide a default container_env_file file? Does it push to examples repo?

modules/universal/scalardl/main.tf
ymorimo: Thanks! Updated in 0c09184.
We may create a separate script, but I'm not going to do further since this PR is not for refactoring this part.

modules/universal/scalardl/main.tf
ymorimo: Added a missing trigger in 199043a. Thank you for pointing it out.

ymorimo:
> Do we provide a default container_env_file file? Does it push to examples repo?

I'm adding example env files in https://github.com/scalar-labs/scalar-terraform-examples/pull/17/files

modules/universal/scalardl/main.tf
ymorimo: Let me think about triggers.

modules/universal/scalardl/main.tf
tei-k: Well, just for bash.
But I think `[[ ` is a bashism and not safer too. (`==` also does not work in some shells )
https://wiki.ubuntu.com/DashAsBinSh#My_production_system_has_broken_and_I_just_want_to_get_it_back_up.21

modules/aws/scalardl/locals.tf
feeblefakie: I think they can be shared.

modules/azure/scalardl/locals.tf
```diff
@@ -14,7 +14,7 @@ locals {
   internal_domain  = var.network.internal_domain
 
   triggers = [
-    local.scalardl.database == "cassandra" && var.cassandra.start_on_initial_boot ? var.cassandra.provision_ids : var.network.bastion_provision_id
+    length(var.cassandra.provision_ids) > 0 && var.cassandra.start_on_initial_boot ? var.cassandra.provision_ids : var.network.bastion_provision_id
```
feeblefakie, line 17: Not directly related to this PR, but can it depend on `cassandra` variables?
What would happen if we use Cosmos DB as a backend database instead of Cassandra?

modules/universal/scalardl/main.tf
feeblefakie: @ymorimo Are you still thinking about something for this?

feeblefakie (commented):
Overall looking good to me. Left some questions.

modules/universal/scalardl/main.tf
ymorimo: Updated to use `[` instead of `[[`.

modules/universal/scalardl/main.tf
ymorimo: Added triggers to the `scalardl_container_env_file_push` resource.

modules/azure/scalardl/locals.tf
ymorimo: I found that it fails to deploy scalardl on Azure. I'll fix the problem.

ymorimo: The map var `var.cassandra` is always defined in `vars.tf`, and `var.cassandra.provision_ids` is expected to be passed from [the examples](https://github.com/scalar-labs/scalar-terraform-examples/blob/main/azure/scalardl/locals.tf#L28) even if we don't use Cassandra. But I updated this line to make it safer to access `var.cassandra.provision_ids` in bdcc65d.

modules/aws/scalardl/locals.tf
```diff
@@ -109,7 +105,7 @@ locals {
 
 locals {
   triggers = [
-    local.scalardl.database == "cassandra" && var.cassandra.start_on_initial_boot ? var.cassandra.provision_ids : var.network.bastion_provision_id
+    length(lookup(var.cassandra, "provision_ids", "")) > 0 && var.cassandra.start_on_initial_boot ? var.cassandra.provision_ids : var.network.bastion_provision_id
```
tei-k, line 108: If we use `lookup` for `provision_ids`, I think we should use `lookup` for `start_on_initial_boot` too.

However, I don't think this trigger is necessary and I think we can remove it in the future.

ymorimo: Ah, that's true. Fixed in 61ec5d5. Thanks! Let's consider removing this trigger later.

feeblefakie:
@ymorimo Please fix the conflicts. I will take a final look once again afterward.

modules/aws/scalardl/cluster/main.tf
```diff
@@ -54,11 +54,7 @@ module "scalardl_provision" {
...
   scalardl_image_name          = var.scalardl_image_name
   scalardl_image_tag           = var.scalardl_image_tag
+  container_env_file           = var.container_env_file
   internal_domain              = var.internal_domain
-  database                     = var.database
-  database_contact_points      = var.database_contact_points
-  database_contact_port        = var.database_contact_port
-  database_username            = var.database_username
-  database_password            = var.database_password
   cassandra_replication_factor = var.cassandra_replication_factor
```
feeblefakie, line 59: It should be done in another PR, but let's make it Cassandra-independent.
Maybe introducing an env file for schema-loader?

modules/universal/scalardl/main.tf
```diff
@@ -192,14 +209,15 @@ resource "null_resource" "scalardl_schema" {
...
-        "docker run --rm ${var.schema_loader_image} -h ${var.database_contact_points} -u ${var.database_username} -p ${var.database_password}",
-        var.database == "cassandra" ? "--cassandra -P ${var.database_contact_port} -n NetworkTopologyStrategy -R ${var.cassandra_replication_factor}" :
-        var.database == "dynamo" ? "--dynamo --region ${var.database_contact_points}" :
-        var.database == "cosmos" ? "--cosmos -h ${var.database_contact_points} -p ${var.database_password}" :
-        ""
-      )
+      "set -x",
+      # Get the SCALAR_DB_* variables from the env file for docker-compose
+      "eval `grep '^SCALAR_DB_' $HOME/provision/container.env`",
+      "cmd=\"docker run --rm ${var.schema_loader_image} -u $SCALAR_DB_USERNAME -p $SCALAR_DB_PASSWORD\"",
```
feeblefakie, line 215: It can be done in another PR, but would it be better to make these lines a script to make the terraform scripts independent from the databases?

feeblefakie (approved):
LGTM! Thank you!
Left some comments for the next improvements.

tei-k (approved):
LGTM!

modules/aws/scalardl/cluster/main.tf
ymorimo: Yes, I will think about how we can remove this variable later.

modules/universal/scalardl/main.tf
ymorimo: Yes, probably we should refactor this part along with the options of `scalardl-schema-loader` image.

PR #294, opened 2021-03-15
Use fixed numbers (50051 and 50052) for all ports used by Envoy and Scalar DL

ymorimo:
https://scalar-labs.atlassian.net/browse/DLT-8326

This is not directly related to Scalar Auditor, but it simplifies the parameters passed to both the `envoy` and `scalardl` modules.
The numbers `50051` and `50052` are always used for communication in LB-Envoy-Scalar DL. These numbers are hard-coded and the variables have been removed.

Now the users can change only the listen port of the load balancer in front of Envoy.


modules/aws/scalardl/locals.tf
```diff
@@ -30,10 +30,8 @@ locals {
     green_image_tag              = "2.1.0"
     green_image_name             = "ghcr.io/scalar-labs/scalar-ledger"
     green_discoverable_by_envoy  = false
-    target_port                  = 50051
-    privileged_target_port       = 50052
-    listen_port                  = 50051
-    privileged_listen_port       = 50052
```
ymorimo, line 36: `listen_port` and `privileged_listen_port` have been never used.

ymorimo, line 36: `listen_port` and `privileged_listen_port` (for scalardl) were not used anywhere.

modules/aws/scalardl/locals.tf
```diff
@@ -72,8 +70,8 @@ locals {
     resource_type             = "t3.medium"
     resource_count            = 3
     resource_root_volume_size = 64
-    target_port               = 50051
-    privileged_target_port    = 50052
+    target_port               = local.scalardl.port
+    privileged_target_port    = local.scalardl.privileged_port
```
ymorimo, line 74: The target ports of Envoy should be the same as Scalar DL ports.

modules/universal/scalardl/provision/docker-compose.yml
```diff
@@ -3,14 +3,14 @@ version: '3.7'
 services:
   scalar:
     image: ${SCALAR_IMAGE}
+    ports:
+      - "${SCALAR_PORT}:${SCALAR_PORT}"
+      - "${SCALAR_PRIVILEGED_PORT}:${SCALAR_PRIVILEGED_PORT}"
+      - "50053:50053"
```
ymorimo, line 9: @feeblefakie Only the admin port is still hard-coded. Should this setting also be customizable?

feeblefakie: I think it would be better if it is parameterized.

modules/universal/envoy/main.tf
```diff
@@ -99,6 +99,8 @@ resource "null_resource" "envoy_container" {
       "cd $HOME/provision",
       "export internal_domain=${var.internal_domain}",
       "export envoy_tls=${var.envoy_tls}",
+      "export port=${var.envoy_port}",
```
feeblefakie, line 102: Sorry, I'm confused.
Setting `var.envoy_port` to both `port` and `ENVOY_PORT`?
Then, what is the point of having `port` and `ENVOY_PORT`?

feeblefakie (commented):
Overall looking good. Left a question and a comment.

modules/universal/envoy/main.tf
ymorimo: `port` is for `j2` command to create the `envoy.yaml` file. It is set in the shell session here, but it is not saved.
`ENVOY_PORT` works in the same way as other variables like `ENVOY_IMAGE`. It is saved to the `env` file and it should be sourced every time `docker-compose` is run.

modules/universal/envoy/main.tf
```diff
@@ -99,6 +99,8 @@ resource "null_resource" "envoy_container" {
...
+      "export privileged_port=${var.envoy_privileged_port}",
       "j2 ./envoy.yaml.j2 > ./envoy.yaml",
       "echo export ENVOY_CERT_AUTO_GEN=${var.envoy_cert_auto_gen}",
```
feeblefakie, line 105: I don't think it's related to this PR, but is it OK not to save this to the `env`?

modules/universal/envoy/main.tf
feeblefakie: Ah, got it. Thanks.

feeblefakie (commented):
Left an additional comment.

modules/universal/scalardl/provision/docker-compose.yml
ymorimo: I will update this PR.

modules/universal/envoy/main.tf
ymorimo: Oh, it seems that this line was intended to take effect in [`auto_gen_cert.sh`](https://github.com/scalar-labs/scalar-terraform/blob/master/modules/universal/envoy/provision/auto_gen_cert.sh), but the variable referenced in that script is wrong. I'd like to address this problem in another PR.

modules/aws/scalardl/locals.tf
tei-k: I don't think these two ports are necessarily the same.

I think it would be better to simplify the customizable ports to just envoy's listen_port(privileged_port) and scalardl's port (privileged_port).

`envoy.listen_port` → <envoy lb> → `envoy.target_port`  → `envoy.target_port:50051` (envoy docker-compose) → 50051(envoy listen_port in container) → `scalardl.port` (envoy.yaml) → `scalardl.port:50051` (scalardl docker-compose) → 50051 scalardl lisetn_port in container

① envoy.listen_port for lb
② envoy.target_port
③ envoy listen_port in container (50051)
④ scalardl.port
⑤ scalardl listen_port in container

modules/aws/scalardl/scalardl.tf
```diff
@@ -92,8 +96,8 @@ resource "aws_security_group_rule" "scalardl_target_port" {
   count = local.scalardl.green_resource_count > 0 || local.scalardl.blue_resource_count > 0 ? 1 : 0
 
   type        = "ingress"
-  from_port   = local.scalardl.target_port
-  to_port     = local.scalardl.target_port
+  from_port   = local.scalardl.port
+  to_port     = local.scalardl.port
```
tei-k, line 100: Do we need to change the resource name as `scalardl_port`? 

`resource "aws_security_group_rule" "scalardl_target_port"`

modules/universal/scalardl/provision/docker-compose.yml
```diff
@@ -3,14 +3,14 @@ version: '3.7'
 services:
   scalar:
     image: ${SCALAR_IMAGE}
+    ports:
+      - "${SCALAR_PORT}:${SCALAR_PORT}"
+      - "${SCALAR_PRIVILEGED_PORT}:${SCALAR_PRIVILEGED_PORT}"
```
tei-k, line 8: I think it would be better to separate the two ports of port mapping.

If the `SCALAR_PORT` is customized by the user, it may not work properly!

modules/aws/scalardl/scalardl.tf
ymorimo: Thank you! Updated the description and the resource name.
Considering the existing environments, the resource name was left as is, but it is ok to be recreated.

modules/universal/scalardl/provision/docker-compose.yml
ymorimo: Let's talk about it tomorrow.

feeblefakie (approved):
LGTM! Thank you!

tei-k (approved):
LGTM!

PR #296, opened 2021-03-22
Set disk scheduler for cassandra

tei-k:
# Description

https://scalar-labs.atlassian.net/browse/DLT-7211

# Done
Fix `scheduler` to `none` for `Azure`.
📝 `AWS` can't change it.


provision/ansible/playbooks/roles/cassandra/files/disk_io_params.service
```diff
@@ -0,0 +1,11 @@
+[Unit]
+Description=Disk io parameters
+After=cloud-final.service
+Wants=cloud-final.service
+
+[Service]
+Type=simple
```
ymorimo, line 7: Probably `oneshot` is the right option?
https://github.com/scalar-labs/scalar-terraform/blob/74426c25952c5122c18c90b649e3fcd4d7e7c1ea/provision/ansible/playbooks/roles/cassandra/files/drivemount.service

```suggestion
Type=oneshot
RemainAfterExit=yes
```

tei-k: Good point! 👍 
Fixed in 56cc0ad

ymorimo: I think `RemainAfterExit=yes` is also needed. Otherwise, it seems that the state will never become "active".

https://www.freedesktop.org/software/systemd/man/systemd.service.html
> Note that if this option is used without RemainAfterExit= the service will never enter "active" unit state, but directly transition from "activating" to "deactivating" or "dead" since no process is configured that shall run continuously.

tei-k: I don't think it's necessary, Because it's not a resident process(`inactive`).

```
[centos@cassandra-1 ~]$ systemctl status disk_io_params
● disk_io_params.service - Disk io parameters
   Loaded: loaded (/etc/systemd/system/disk_io_params.service; enabled; vendor preset: disabled)
   Active: inactive (dead) since Tue 2021-03-23 08:37:04 UTC; 21s ago
  Process: 14607 ExecStart=/usr/local/bin/disk_io_params.sh (code=exited, status=0/SUCCESS)
 Main PID: 14607 (code=exited, status=0/SUCCESS)
```

```
RemainAfterExit=

Takes a boolean value that specifies whether the service shall be considered active even when all its processes exited. Defaults to no.
```

ymorimo: Right, but the `drivemount.service` specifies `RemainAfterExit=yes` with `Type=oneshot`. What do you think the difference is?

provision/ansible/playbooks/roles/cassandra/defaults/main.yml
```diff
@@ -14,3 +14,10 @@ cassandra_dc_id: dc1
 cassandra_rack_id: rack1
 endpoint_snitch: GossipingPropertyFileSnitch
 vm_max_map_count: "1048575"
+
+disk_io_params:
+  - { name: scheduler, value: none }
+disk_io_devices:
+  - sda
```
feeblefakie, line 21: Is it for Azure?  It's always sda, sdb, and sdc?
If it is for only Azure, it should be noted as so.

feeblefakie (commented):
Left one question.

provision/ansible/playbooks/roles/cassandra/files/disk_io_params.service
tei-k: I need to update disk parameters by starting the service.
When set to `RemainAfterExit=yes` , it will not be possible to update after the first service startup. (difference with drivemount)

https://github.com/scalar-labs/scalar-terraform/pull/296/files#diff-610e26d306e2b7c67547e7b09f5681cfc6a10c59028ddad5c62b063af3a70522R14-R19

provision/ansible/playbooks/roles/cassandra/defaults/main.yml
tei-k: Yes, only for Azure (if there is no file on the device, it is ignored in AWS)

https://github.com/scalar-labs/scalar-terraform/pull/296/files#diff-622a8b817c51db77fcaa4f894b48e1e3e8a05b3af65e8eef3465707945731732R5

feeblefakie: Ok, then, please note so (and why we don't do this for AWS). Otherwise, others who don't know our discussion has no way to understand this code.

provision/ansible/playbooks/roles/cassandra/files/disk_io_params.service
ymorimo: When `RemainAfterExit=yes` is set, isn't it possible to run the service again by `systemctl restart disk_io_params` as needed?

ymorimo: I've got the situation. Thanks!

ymorimo (approved):
LGTM!

provision/ansible/playbooks/roles/cassandra/defaults/main.yml
feeblefakie: @tei-k Overall looking good except for this. So please note why it is only for Azure.

tei-k: Added comment in c7ae6ce.

ymorimo (approved):
LGTM!

provision/ansible/playbooks/roles/cassandra/defaults/main.yml
```diff
@@ -14,3 +14,12 @@ cassandra_dc_id: dc1
 cassandra_rack_id: rack1
 endpoint_snitch: GossipingPropertyFileSnitch
 vm_max_map_count: "1048575"
+
+disk_io_params:
+  - { name: scheduler, value: none }
+
+# Only the list of disk devices in the Azure. Since disk io parameters in the AWS instances cannot be changed.
```
feeblefakie, line 21:
```suggestion
# The following list of devices is for only Azure since the kernel parameters of AWS instances cannot be changed.
```

feeblefakie (approved):
LGTM!

PR #298, opened 2021-03-23
Update required Terraform version to 0.14 in README

ymorimo:
The required Terraform version in the README is still 0.12 while the [Getting Started](https://github.com/scalar-labs/scalar-terraform/blob/97b2bb2f069e3694322607d89e35af2e832ae9a1/docs/GettingStarted.md) doc requests 0.14.

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,7 +7,7 @@ Scalar Terraform is a set of terraform modules and provisioing scritps that can 
 
 ## Requirements
 
-* [Terraform >= 0.12.x](https://www.terraform.io/downloads.html)
+* [Terraform >= 0.14.x](https://www.terraform.io/downloads.html)
 * [Ansible >= 2.8.x](https://docs.ansible.com/ansible/latest/installation_guide/intro_installation.html)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -7,7 +7,7 @@ Scalar Terraform is a set of terraform modules and provisioing scritps that can
 
 ## Requirements
 
-* [Terraform >= 0.12.x](https://www.terraform.io/downloads.html)
+* [Terraform >= 0.14.x](https://www.terraform.io/downloads.html)
```
tei-k, line 10: Maybe better to write like this ?
```
[Terraform ~> 0.14.7]
```

https://github.com/scalar-labs/scalar-terraform/blob/master/modules/aws/network/README.md

ymorimo: The `~>` is a common notation but it's not very clear to humans. And I don't think it makes much sense to show the version to the bugfix number here. If we update the notation, I think the following would be ok for README.

```
Terraform 0.14
Ansible 2.8
```

feeblefakie (approved):
LGTM!

PR #299, opened 2021-03-26
Fix Envoy auto_gen_cert script

ymorimo:
This is the fix for the problem pointed out here. https://github.com/scalar-labs/scalar-terraform/pull/294#discussion_r595721629

Even though `envoy.cert_auto_gen` was set to `true` (it was the default), `auto_gen_cert.sh` was not working. And since the variable `envoy.tls` is `false` by default, running `auto_gen_cert.sh` by default didn't make sense. So I modified the default of that variable to `false`.
I wonder if this script is really needed because it creates a self-signed certificate, but this is the bugfix anyway.

### DIFF
--- a/modules/aws/scalardl/locals.tf
+++ b/modules/aws/scalardl/locals.tf
@@ -79,7 +79,7 @@ locals {
     tag                       = "v1.14.1"
     image                     = "envoyproxy/envoy"
     tls                       = false
-    cert_auto_gen             = true
+    cert_auto_gen             = false
     custom_config_path        = ""
   }
 }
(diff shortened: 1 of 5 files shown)


feeblefakie (dismissed):
LGTM!

tei-k:
I think it would be better to update the following default value to `false`

https://github.com/scalar-labs/scalar-terraform/blob/f02ce20b45922fc5f6ff814426d7a2a85829bf6b/modules/universal/envoy/vars.tf#L57-L60

ymorimo:
@tei-k Thanks! Updated.

tei-k (approved):
LGTM!

PR #303, opened 2021-04-07
Remove triggers of cassandra from scalardl module

ymorimo:
https://scalar-labs.atlassian.net/browse/DLT-8836

The triggers for `null_resource` resources in the scalardl module are no longer needed. Accordingly, the outputs from the cassandra module are also removed.

### DIFF
--- a/modules/aws/cassandra/README.md
+++ b/modules/aws/cassandra/README.md
@@ -38,11 +38,9 @@ The Cassandra AWS module deploys a Cassandra cluster tuned for a Scalar DL envir
 | cassandra_host_ids | A list of host IDs for the Cassandra cluster. |
 | cassandra_host_ips | A list of host IP addresess for the Cassandra cluster. |
 | cassandra_hosts | A list of dns urls to access the Cassandra cluster. |
-| cassandra_provision_ids | The IDs of the provisioning step. |
 | cassandra_resource_count | The number of Cassandra nodes to create. |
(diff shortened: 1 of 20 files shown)


tei-k:
Also need to update the following docs.

https://github.com/scalar-labs/scalar-terraform/blob/master/modules/aws/cassandra/README.md
https://github.com/scalar-labs/scalar-terraform/blob/master/modules/azure/cassandra/README.md

tei-k (approved):
LGTM!

feeblefakie (approved):
LGTM!

@ymorimo @TakashiChikata 
Just a question.
I think it's fine if we deploy by manually since we manage the dependency and wait for C* to start before deploying scalardl.
But, isn't it OK for terratest? (Not sure how scalardl waits for C* to be ready in terratest.)

ymorimo:
@feeblefakie Now there's no dependency between the `cassandra` and `scalardl`  modules, so they can be deployed independently and there's no need to wait for `cassandra` to be deployed when `scalardl` is being deployed In the deployment phase.

tei-k:
>But, isn't it OK for terratest?

The process of 3 and 4 together may take about 5 minutes, so the terratest is OK.

1. Create cassandra resources
2. Provision with ansible for Cassandra (auto start cassandra)
3. Create scalardl resources
4. provision with ansible for scalardl
5. Create scalardl schema resource
6. Create scalardl container resource

PR #305, opened 2021-04-16
Fix the schema-loading command for Cosmos DB

ymorimo:
When we use Cosmos DB, `SCALAR_DB_USERNAME` is not used but the option `-u` was specified in the schema loading command.

If we use a `scalardl_conatainer.env` like below with an empty `SCALAR_DB_USERNAME=` line,
```
SCALAR_DB_STORAGE=cosmos
SCALAR_DB_CONTACT_POINTS=https://xxxxx-cosmosdb.documents.azure.com:443/
SCALAR_DB_CONTACT_PORT=
SCALAR_DB_USERNAME=
SCALAR_DB_PASSWORD=the-cosmos-db-key
```
the following command is run and it fails due to no argument for `-u`.
```
docker run --rm ghcr.io/scalar-labs/scalardl-schema-loader:1.3.0 -u -p the-cosmos-db-key --cosmos -h https://xxxxx-cosmosdb.documents.azure.com:443/
Exception in thread "main" java.lang.NullPointerException: 'key' cannot be null.
...
```

[As we discussed](https://github.com/scalar-labs/scalar-terraform/pull/293#discussion_r611266780), we need to consider refactoring the schema loading part, but this is a quick fix.

### DIFF
--- a/modules/universal/scalardl/main.tf
+++ b/modules/universal/scalardl/main.tf
@@ -212,9 +212,9 @@ resource "null_resource" "scalardl_schema" {
       "set -x",
       # Get the SCALAR_DB_* variables from the env file for docker-compose
       "eval `grep '^SCALAR_DB_' $HOME/provision/container.env`",
-      "cmd=\"docker run --rm ${var.schema_loader_image} -u $SCALAR_DB_USERNAME -p $SCALAR_DB_PASSWORD\"",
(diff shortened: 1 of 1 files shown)


tei-k (approved):
LGTM!

feeblefakie (approved):
LGTM!

PR #306, opened 2021-04-26
Update Scalar DL node replacement command

scalar-boney:
Based on NodeReplacement doc, Scalar DL node replacement failed in both AWS and Azure PSIM environments.

```
module.scalardl.module.scalardl_blue.module.scalardl_provision.null_resource.scalardl_load[0] (remote-exec): 52d41b419360: Loading layer  5.632kB/5.632kB
module.scalardl.module.scalardl_blue.module.scalardl_provision.null_resource.scalardl_load[0] (remote-exec): 


module.scalardl.module.scalardl_blue.module.scalardl_provision.null_resource.scalardl_load[0] (remote-exec): 2e790ed37b7e: Loading layer  3.072kB/3.072kB
module.scalardl.module.scalardl_blue.module.scalardl_provision.null_resource.scalardl_load[0] (remote-exec): 
module.scalardl.module.scalardl_blue.module.scalardl_provision.null_resource.scalardl_load[0] (remote-exec): Loaded image: ghcr.io/scalar-labs/scalar-ledger:2.1.0
module.scalardl.module.scalardl_blue.module.scalardl_provision.null_resource.scalardl_load[0]: Still creating... [1m10s elapsed]
module.scalardl.module.scalardl_blue.module.scalardl_provision.null_resource.scalardl_load[0]: Creation complete after 1m10s [id=1314021266399735200]

Error: stat .terraform/modules/scalardl/modules/universal/scalardl/scalardl-schema-loader-1.2.0.tar.gz: no such file or directory
```

Ref:- https://scalar-labs.atlassian.net/browse/DLT-8757
https://scalar-labs.atlassian.net/browse/DLT-8735

PR #307, opened 2021-04-26
Add file check for create scalardl image file

tei-k:
# Description

Errors occur when running `taint` on a different machine than the one used to build the environment.

ref: #306


tei-k:
Recreated PR #310

PR #314, opened 2021-05-24
Add create release draft workflow

tei-k:
# Description

new release tool.
https://github.com/release-drafter/release-drafter

ref: https://github.com/twbs/bootstrap/blob/main/.github/release-drafter.yml

# Confirm
Screen Shot 2021-05-24 at 17 21 52

# How to use
- Create PR with label
  - for resolve version (`major`, `minor`, `patch`)
  - for release notes (`enhancement`, `bugfix`, `documentation`, `improvement`)
- Merge to master
- Create/update release draft by github actions


.github/release-drafter.yml
```diff
@@ -0,0 +1,32 @@
...
+tag-template: 'v$RESOLVED_VERSION'
+prerelease: true
+exclude-labels:
+  - 'skip-changelog'
+categories:
+  - title: 'Enhancements'
+    labels:
+      - 'feature'
+      - 'internal'
+      - 'enhancement'
```
ymorimo, line 11: I don't think `feature` and `internal` labels are necessary to be listed since we don't use them.

.github/workflows/release_draft.yml
```diff
@@ -0,0 +1,14 @@
+name: Release Drafter
```
ymorimo, line 1: Isn't it normal to name this file `.github/workflows/release-drafter.yml` as long as I see the examples?

ymorimo (commented):
I didn't know about this tool, but it's a good one!
Left two minor comments.

.github/release-drafter.yml
tei-k: Thanks! These are from old release tool. deleted in 96c3282.

.github/workflows/release_draft.yml
tei-k: Renamed file name to `release_drafter.yml` in 9ee93cc. (Used underscore to match other workflow files.)

ymorimo (approved):
LGTM!

feeblefakie (approved):
LGTM! Thanks!

PR #315, opened 2021-05-25
Add scalardl node replacement postmortem report

scalar-boney:
Scalardl node replacement failed in the Azure PSIM environment while trying to replace the accidentally deleted node.

PR #316, opened 2021-05-27
Fix terraform refresh in docs

tei-k:
# Description
https://github.com/scalar-labs/scalar-terraform/pull/315#discussion_r639362264

`refresh` command is deprecated. Fix to use `apply` instead of `refresh`.

https://www.terraform.io/docs/cli/commands/refresh.html
```
Warning: This command is deprecated, because its default behavior is unsafe if you have misconfigured credentials for any of your providers. See below for more information and recommended alternatives.
```

### DIFF
--- a/docs/HowToExpandAzureManagedDisk.md
+++ b/docs/HowToExpandAzureManagedDisk.md
@@ -75,8 +75,8 @@ Second, do `terraform plan` to check if Cloud resources' states are the same as 
 $ terraform plan -var-file=your.tfvars
 ```
 
-Third, do `terraform refresh` to update tfstate.
+Third, do `terraform apply` to update tfstate.
 
 ```console
-$ terraform refresh -var-file=your.tfvars
+$ terraform apply -var-file=your.tfvars
 ```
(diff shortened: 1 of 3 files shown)


ymorimo:
According to the document, isn't `terraform apply -refresh-only` a proper alternative in some cases? (Terraform >= v0.15.4 is required.)

ymorimo:
In other words, `terraform refresh` is not deprecated as long as we don't upgrade terraform to v0.15.4, isn't it?

tei-k:
>According to the document, isn't terraform apply -refresh-only a proper alternative in some cases? (Terraform >= v0.15.4 is required.)

Yes, We can use it after upgrading terraform version.

>In other words, terraform refresh is not deprecated as long as we don't upgrade terraform to v0.15.4, isn't it?

No, My understanding is that `terraform refresh` is not recommended in all versions.
Because it provides no opportunity to review the effects of the operation before updating the state.

feeblefakie (approved):
LGTM! Thanks!

ymorimo (approved):
LGTM!

scalar-boney (approved):
LGTM!

PR #319, opened 2021-06-02
Fix to use prefix version for azure k8s module

tei-k:
# Description

Due to the fast-moving nature of AKS, we recommend using the latest version, So it is better to specify a prefix version like aws.

https://scalar-labs.atlassian.net/browse/DLT-9166

# Done
- Use the latest version searched by prefix.
- Add [orchestrator_version](https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs/resources/kubernetes_cluster_node_pool#orchestrator_version)
```
orchestrator_version - (Optional) Version of Kubernetes used for the Agents. If not specified, the latest recommended version will be used at provisioning time (but won't auto-upgrade)
```


ymorimo (approved):
LGTM!

feeblefakie (approved):
LGTM! Thank you!

PR #326, opened 2021-07-29
Refactoring cosmosdb module

tei-k:
# Description

https://scalar-labs.atlassian.net/browse/DLT-9322

# Done

- Add `use_kubernetes` flag for allow access from k8s node.
- Upgrade azure provider
- Add `allowed_cidrs` for firewall.

### DIFF
--- a/modules/azure/cosmosdb/main.tf
+++ b/modules/azure/cosmosdb/main.tf
@@ -10,10 +10,10 @@ resource "azurerm_cosmosdb_account" "db" {
   is_virtual_network_filter_enabled = true
 
   dynamic "virtual_network_rule" {
-    for_each = var.allowed_subnet_ids
-    iterator = subnet_id
+    for_each = toset(var.allowed_subnet_ids)
     content {
-      id = subnet_id.value
+      id                                   = each.key
+      ignore_missing_vnet_service_endpoint = false
     }
   }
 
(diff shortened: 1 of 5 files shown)


modules/azure/cosmosdb/main.tf
```diff
@@ -7,16 +7,18 @@ resource "azurerm_cosmosdb_account" "db" {
 
   enable_automatic_failover = false
 
-  is_virtual_network_filter_enabled = true
+  is_virtual_network_filter_enabled = var.is_virtual_network_filter_enabled
```
ymorimo, line 10: Just a question. Is it needed to make this variable? Is there any case when we need to change this to `false`?

tei-k: Good point! I added it because it might be useful for debugging from local or antoher networks.

ymorimo: Ok. Then, `is_virtual_network_filter_enabled` is the same name as the `azurerm_cosmosdb_account`'s parameter, but as a scalar-terraform module, the name should probably be like `enable_virtual_network_filter`?

tei-k: @ymorimo 

The name `enable_virtual_network_filter ` looks good. Thank you! Fixed in 6eb29f4

ymorimo (approved):
LGTM!

feeblefakie (approved):
LGTM! Thank you!

PR #331, opened 2021-10-26
Modify bastion default resource type for Azure

miseyu:
## Summary
- Changed the default instance type because there were cases where the memory of `bastion` was not enough when building infrastructure.

### DIFF
--- a/modules/azure/network/locals.tf
+++ b/modules/azure/network/locals.tf
@@ -1,7 +1,7 @@
 ### default
 locals {
   network_default = {
-    bastion_resource_type                 = "Standard_B1s"
+    bastion_resource_type                 = "Standard_B2s"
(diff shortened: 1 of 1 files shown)


MilanCorreggio (commented):
So this Ansible error :`A worker was found in a dead state` was due to a lack of memory from the bastion ?
if so, because this problem did not appear today shall we wait before upgrading the bastion type ?
I also saw that we can run ansible using `-fx` to limit this but it might affect performance.
```
-f NUM, --forks=NUM
Level of parallelism. NUM is specified as an integer, the default is 5.
```

MilanCorreggio (approved):

feeblefakie (approved):
LGTM! Thank you!

